PR #321, opened 2016-10-03
Added MonteVar setters where appropriate

ddj116:
Added set_unit mechanism to MonteVar baseclass. Added MonteVar's derived
classes setters where appropriate to facilitate changing inputs after
the instance of MonteVar has already been constructed. Fleshed out
MonteVarFile destructor to remove memory leaks.


ddj116:
Let me know if y'all have feedback, I'm deliberately waiting until this gets accepted in the master branch before "rolling it out" in our version of Trick so I can avoid any binary rebuilds/commits if there's needed changes.

PR #409, opened 2017-03-31
Reduced size of generated argument list for simulation library creation

pbielski:
Change the way the simulation object library is generated to eliminate long paths at the beginning of each file going into the argument list for the ar command. The make command on macOS is sensitive to the length of the argument list, and the ar command would fail when the argument list gets north of 200,000 characters.


alexlin0:
Done!

Issue #416, opened 2017-05-07
Trick doesn't compile clean with clang++ -Wdocumentation


DavidHammen:
In fact, it spews.

There are a number of inconsistencies between the doxygen comments and the code in Trick. Inconsistencies between comment and code are indications that both the comment and the code are incorrect. Having a tool that checks for such inconsistencies is a good thing. The `-Wdocumentation` option in clang++ does just that, checking for consistency between the doxygen comments and the code.

For example, the following will receive a warning that the documented parameter `foobar` is not in the prototype, and that it's invalid to document the return value for a function that returns `void`.

    /**
     * Foo the subject object.
     * \param foobar  Object to be fooed.
     * \return True if the object was fooed, false otherwise.
     */
    void foo (SomeType bar);

PR #417, opened 2017-05-07
Hammen doxygen comments

DavidHammen:
Issued against issue #416. The pull request is against tag 15.2.3. (Sorry Alex; that's what we're using as a baseline.)

And no, I don't want to merge 51 commits into `nasa:master`. My changes are against 15.2.3. Just reject this pull request. I'll issue another one.

Stet that. **Love** github. These changes are against the 15.2.3 tag. I most likely dyslexically typed 15.3.2, and github most likely said that that doesn't exist, it would use `master`. And now I can't create a new pull request. As I wrote earlier, I just **love** github.


alexlin0:
OK, rejecting as requested.

PR #434, opened 2017-05-22
Capped the maximum throttle value to 1000.

CTLaChance:
Entering extremely large values into the maximum value field of the throttle GUI caused the GUI to freeze. Imposing a cap of 1000 to the maximum value prevents the GUI from freezing.


dbankieris:
Nobody uses that slider. Get rid of it!

CTLaChance:
That's a potential solution. What do you guys think?

CTLaChance:
Merged the branches and locked this pull request. Issue closed.

Issue #391, opened 2017-03-14
TrickView Doesn't Support NAN Floats


doylewb:
When a float value is NAN Trick View breaks and shows 'Unavailable' instead of 'NAN' or the hexadecimal equivalent. This occurred on Trick 13.5.0.

CTLaChance:
All done! This is now part of the master branch.

PR #437, opened 2017-06-06
+-NaN and +-Infinity are now correctly represented in Trick View.

CTLaChance:
This is a bug fix referencing issue #391.

### DIFF
--- a/trick_source/java/src/trick/common/utils/vs/VSDouble.java
+++ b/trick_source/java/src/trick/common/utils/vs/VSDouble.java
@@ -12,24 +12,7 @@ public class VSDouble extends VSValue<Double> {
 
     @Override
     public void fromVariableServer(String string) {
-        // CTLaChance
-        // JUN 07 2017
-        /*
-        Converts the C representation of +- infinity and +- not-a-number to the Java representation
-        so that Double.parseDouble() can correctly assign those values to the variable value.
-        -NaN does not exist in Java and is therefore converted to NaN.
-        */
-        if(string.equals("inf")) {
-            string = "Infinity";
-        }
-        else if(string.equals("-inf")) {
-            string = "-Infinity";
-        }
-        else if(string.equals("-nan") || string.equals("nan")) {
-            string = "NaN";
-        }
-
-        setValue(Double.parseDouble(string.trim()));
+        setValue(Double.parseDouble(handleUndefinedValues(string).trim()));
     }
 
     @Override
--- a/trick_source/java/src/trick/common/utils/vs/VSDouble.java
+++ b/trick_source/java/src/trick/common/utils/vs/VSDouble.java
@@ -12,6 +12,23 @@ public class VSDouble extends VSValue<Double> {
 
     @Override
     public void fromVariableServer(String string) {
+        // CTLaChance
+        // JUN 07 2017
+        /*
+        Converts the C representation of +- infinity and +- not-a-number to the Java representation
+        so that Double.parseDouble() can correctly assign those values to the variable value.
+        -NaN does not exist in Java and is therefore converted to NaN.
+        */
+        if(string.equals("inf")) {
+            string = "Infinity";
+        }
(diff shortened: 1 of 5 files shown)


dbankieris:
**Problems**
1. This fix is too broad. `NaN` and `Infinity` can only occur for floating point numbers, so logic to handle them should be restricted to `trick_source/java/src/trick/common/utils/vs/VSFloat.java` and `VSDouble.java`.
2. `VSVariable.value` is meant to reflect the latest (valid) value from the variable server. This fix leaves the current value unchanged, whereas we should actually set it to `NaN` or `Infinity`.
3. `NaN` and `Infinity` are valid values. Introducing a new value for the `State` enum to represent them suggests they are neither valid nor invalid (nor unknown), which is incorrect.

**Suggestions**
 `VSFloat.fromVariableServer` calls `Float.parseFloat`, which is [perfectly capable](https://docs.oracle.com/javase/8/docs/api/java/lang/Float.html#valueOf-java.lang.String-) of parsing `NaN` and `Infinity`. Unfortunately, Java and C++ (and therefore Trick View and the variable server) don't agree about the string representations of these values:

| Value |   Java   | C++ |
|:-----:|:--------:|:---:|
|  0/0  |    NaN   | nan |
|   ∞   | Infinity | inf |

Possible solutions:
1. Modify the variable server to use Java string representations for `NaN` and `Infinity`.
2. Modify `VSFloat.fromVariableServer` to detect and convert C++ representations to Java representations before calling `Float.parseFloat`. Repeat for `VSDouble`.

I lean towards the second option. Changing the variable server could break clients that rely on the C++ representations (although I doubt there are any). Changing Trick View won't impact anyone.

CTLaChance:
On it.

CTLaChance:
Check out these new changes I made.

trick_source/java/src/trick/common/utils/vs/VSDouble.java
```diff
@@ -12,6 +12,23 @@ public VSDouble(double value) {
...
+        so that Double.parseDouble() can correctly assign those values to the variable value.
+        -NaN does not exist in Java and is therefore converted to NaN.
+        */
+        if(string.equals("inf")) {
+            string = "Infinity";
+        }
+        else if(string.equals("-inf")) {
+            string = "-Infinity";
+        }
+        else if(string.equals("-nan") || string.equals("nan")) {
```
dbankieris: `Double.parseDouble` is capable of parsing both `"NaN"` and `"-NaN"`. Although we have observed that it returns `NaN` for both, we should not presume it will always do so. All we want to do here is convert the incoming string to something Java can understand. Convert `"nan"` to `"NaN"` and `"-nan"` to `"-NaN"`, and let Java decide what to do with them.

trick_source/java/src/trick/common/utils/vs/VSDouble.java
```diff
@@ -12,6 +12,23 @@ public VSDouble(double value) {
 
     @Override
     public void fromVariableServer(String string) {
+        // CTLaChance
```
dbankieris: Don't include author/date information. This metadata is handled by git.

trick_source/java/src/trick/common/utils/vs/VSDouble.java
```diff
@@ -12,6 +12,23 @@ public VSDouble(double value) {
...
+        // JUN 07 2017
+        /*
+        Converts the C representation of +- infinity and +- not-a-number to the Java representation
+        so that Double.parseDouble() can correctly assign those values to the variable value.
+        -NaN does not exist in Java and is therefore converted to NaN.
+        */
+        if(string.equals("inf")) {
```
dbankieris: This code is replicated in `VSFloat.fromVariableServer`, and I have a special hatred in my heart for code duplication. I recommend lifting this logic into a standalone protected function in the super class.

CTLaChance:
Got it. Will fix now.

Issue #398, opened 2017-03-22
Monte Carlo: default_port_flag


dbankieris:
`set_listen_device_port` and `set_connection_device_port` set the ports directly in the `TCDevice`s and then set `default_port_flag` to `false` so that `socket_init` doesn't reset them to `0`. Why the convoluted logic? Just initialize them to `0` and set them as desired. Remove `default_port_flag` and `socket_init`.

CTLaChance:
34cc3ae8e5f62ac60ad9d8e1d1a0abf52cdccf7e

Is this what you were suggesting?

CTLaChance:
This issue has been addressed and is now apart of the master branch.

PR #442, opened 2017-06-16
Removed default_port_flag and socket_init. 

CTLaChance:
This is in reference to issue #398.


CTLaChance:
Jenkins is not happy. :(

PR #446, opened 2017-06-21
Replaced socket_init calls with tc_init.

CTLaChance:
Some classes were still calling socket_init which had been removed. Replaced those calls with tc_init.

Make compiled and make test successfully ran all tests. Jenkins should be happy now.

This pull requests references the following issue: #398


CTLaChance:
I think that's everything. . .

PR #454, opened 2017-07-06
ICG: Simplify stl_classes declaration

iamthad:
I think this is easier to read, and having it declared const may result in some performance benefit.


iamthad:
Withdrawing this PR for now, will include the changes here in a larger PR.

Issue #455, opened 2017-07-11
Trick needs a way to display a list of available units


jmpenn:
Since we moved to **udunits**, we don't have a way to display a list of available units that a user can use in their simulation, short of digging into the udunits xml units database. 

The way to find the database is to invoke "udunits2 -h", and then find the default [[ath to the database at the bottom of the printout.

> % udunits2 -h
> Usage:
>     udunits2 -h
>     udunits2 [-A|-L|-U] [-r] [-H have] [-W want] [XML_file]
> 
> where:
>     -A         Use ASCII encoding (default).
>     -L         Use ISO-8859-1 (ISO Latin-1) encoding.
>     -U         Use UTF-8 encoding.
>     -h         Help.  Print this message.
>     -r         Reveal any problems in the database.
>     -H have    Use "have" unit for conversion. Default is reply to prompt.
>     -W want    Use "want" unit for conversion. Empty string requests
>                definition of "have" unit. Default is reply to prompt.
>     XML_file   XML database file. Default is "/usr/local/Cellar/udunits/2.2.24/share/udunits/udunits2.xml".

CTLaChance:
What do you propose? Should there be a "trick-units" perl script in 'bin' that users can run to receive a print-out of all the units they can use? Should trick throw warnings and direct the user to the appropriate documentation when it receives values for units that it doesn't recognize?

Will the list of available units change significantly over time? If not, simply adding the list of available units to the documentation would be a simple and immediate benefit to the user. If so, creating a program to display the available unit types might be a better solution.

alexlin0:
A stand alone script or program would be fine.   Perl would be an excellent choice.  This was the previous output when we ran "ICG -u".  Just reproducing something similar is what we are currently looking for.  And to answer your question, the list of units won't change much.

```
Trick Measurement Units Summary
-------------------------------
                Time:   s min  hr day
Angular Displacement:   r   d  as  am rev
             Voltage:   v
            Amperage: amp
          Resistance: ohm
               Sound:  dB
            Unitless:  -- cnt one mol

English System Units
--------------------
 Linear Displacement:  ft  in  yd  mi  n.m.
                Mass:  sl lbm
               Force:  oz lbf
         Temperature:   R   F
              Energy:   BTU
               Power:   hp
            Pressure:   psi

Metric System Units
-------------------
 Linear Displacement:   m
                Mass:   g  mt
               Force:   N
         Temperature:   C   K
              Energy:   J   TNT
               Power:   W 
[...]

PR #457, opened 2017-07-18
Added a new binary file that will display Trick's available units when called.

CTLaChance:
This pull request references issue #455.


CTLaChance:
Looks like the actual information in the file was incorrect. I was using Trick's old units as the basis. Will update the file and perform a new pull request later.

PR #469, opened 2017-08-09
Fixes #468

### LINKED ISSUE #468: convert_swig mishandles struct templates
convert_swig parses `template <argument,list> struct StructName {...}` as `template <argument,list>` and then `struct StructName {...}` separately.  This results in a `%trick_cast_as` call for `StructName`, generating invalid code.

The fix is easy: Change `class` in `$template_def` to `(class|struct)`.

DavidHammen:
Fixes issue https://github.com/nasa/trick/issues/468.

This is not a complete fix. Template template parameters (e.g., `template< template<class> class T> class Foo ...`) aren't parsed correctly.

### DIFF
--- a/libexec/trick/convert_swig
+++ b/libexec/trick/convert_swig
@@ -90,7 +90,8 @@ my $class_def   = qr/(?:class|struct)\s*                   # keyword class or st
              /sx ;
 my $template_def   = qr/template\s*                        # keyword template
             <[^>]+>\s*                                     # template parameters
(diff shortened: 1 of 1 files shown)


DavidHammen:
Fixes issue https://github.com/nasa/trick/issues/468.

This is not a complete fix. Template template parameters (e.g., `template< template<class> class T> class Foo ...`) aren't parsed correctly.

Issue #316, opened 2016-09-22
MonteVarFile does not support non-numeric values


alexenos:
It would be nice to support string values in the MonteVarFile parsing, otherwise, please explicitly state the limitation in the user's guide.

At the moment the strtold() at MonteVarFile.cpp:31 explicitly prevents the parsing of string values within a MonteVarFile.

PR #488, opened 2017-10-12
Non-numerical values can now be used in Monte Carlo data files.

CTLaChance:
This pull request is in response to #316.

By converting data file traversal from strtold to strtok, it is possible to accept non numerical values in data files. If sending a string, it may be necessary to surround the string with "quotations". Non_numerical values  can also not contain spaces or tabs.

### DIFF
--- a/trick_source/sim_services/MonteCarlo/MonteVarFile.cpp
+++ b/trick_source/sim_services/MonteCarlo/MonteVarFile.cpp
@@ -8,8 +8,7 @@
 #include "trick/message_proto.h"
 #include "trick/exec_proto.h"
 
-Trick::MonteVarFile::MonteVarFile(std::string in_name, std::string in_file_name, unsigned int in_column, std::string in_unit) :
-    input_file_stream(NULL)
+Trick::MonteVarFile::MonteVarFile(std::string in_name, std::string in_file_name, unsigned int in_column, std::string in_unit) : input_file_stream(NULL)
 {
     this->name = in_name;
     this->column = in_column;
@@ -19,7 +18,8 @@ Trick::MonteVarFile::MonteVarFile(std::string in_name, std::string in_file_name,
     buffer = new char[4096];
 }
 
-Trick::MonteVarFile::~MonteVarFile() {
+Trick::MonteVarFile::~MonteVarFile()
+{
     delete input_file_stream;
     delete buffer;
 }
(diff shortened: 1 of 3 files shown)


trick_source/sim_services/MonteCarlo/MonteVarFile.cpp
```diff
@@ -30,64 +30,78 @@ std::string Trick::MonteVarFile::describe_variable()
...
-        char *current_position = buffer;
-        do {
+std::string Trick::MonteVarFile::get_next_value()
+{
+    if (input_file_stream->good())
+    {
+        // Skip the comments and empty lines in the data file.
+        do
+        {
             input_file_stream->getline(buffer, 4096);
```
dbankieris: What happens if there are more than 4096 characters in the line? According to [the documentation](http://www.cplusplus.com/reference/istream/istream/getline/), the first call to `getline` will get the first 4096 characters, and the next call will continue from there, not actually starting on the next line. Although there's some stuff about `sentry` objects at the end that I don't understand. Should probably test this. Either way, this appears to implicitly limit the number of columns that can be used.

Forget all that. Use [`std::getline`](http://en.cppreference.com/w/cpp/string/basic_string/getline) instead. Then you can read into a `std::string` that automatically sizes itself instead of a fixed-size `char` array. No more arbitrary limit. Bonus: it probably simplifies the rest of file. I know you didn't write the original code, and that's not really what your changes are about, but you might as well improve it while you're in here!

trick_source/sim_services/MonteCarlo/MonteVarFile.cpp
```diff
@@ -30,64 +30,78 @@ std::string Trick::MonteVarFile::describe_variable()
...
-            if (input_file_stream->eof()) {
+
+            if(input_file_stream->eof())
+            {
                 input_file_stream->close();
                 return "EOF";
             }
-        } while (file_value == 0 && current_position == buffer && input_file_stream->good());
+        }
+        while(buffer[0] == '#' || buffer[0] == '\0');
```
dbankieris: This is a change to the "comment skipping" logic. Previously, any line that didn't start with a number as determined by `strtold` would be skipped. The new logic is more restrictive, skipping only empty lines and lines that start with `#`. This will break people who rely on the old behavior. Such is the cost of progress!

trick_source/sim_services/MonteCarlo/MonteVarFile.cpp
CTLaChance: +1 Sounds like a good idea. I'll make that change.

trick_source/sim_services/MonteCarlo/MonteVarFile.cpp
CTLaChance: This is correct, however I can not think of a way of skipping comments whilst simultaneously accepting strings as possible input values. The old functionality totally disregarded any value that could not be parsed as a long double. Accepting a string requires a way of skipping comments and null lines.

In the tutorial and user guide, the suggest way of specifying comment lines was to use a '#' character. This change is consistent with that documentation.

CTLaChance:
@dbankieris Here is a simple update to the fix that implements your suggestion.

Issue #432, opened 2017-05-22
MonteVarFile's number of columns equals number of times the files is opened


ddj116:
Recently our group ran into the following error only on particular machines when running a specific monte-carlo run:

```
DIAGNOSTIC: Trick:MonteVarFile the input file "path/to/our/file.txt" failed to open
```
After digging into it a bit, I found that the reason the file failed to open was because we had hit the maximum number of open files for that process on that machine (See #431), in this case the default number is 1024.  We can of course configure our machines to increase this limit, but that number seemed high to me so I started looking a bit closer at `Trick::MonteVarFile::set_file_name` which is where these files are opened.

I noticed that each `MonteVarFile` we create in our input file was resulting in a calls to `Trick::MonteVarFile::set_file_name` which in turn opens the same file many times.  More specifically, the file is opened once for each column of data that the sim is configured to use, at input processor time, and *leaves the file handles open for the instance to use later*.  For example we have one such file that has 190 columns of data, here's what our input file configuration looks like:

```python
var1   = trick.MonteVarFile("simobj.myValue[0]", "Modified_data/monte_carlo/my_mcvar_file.txt", 1)
trick_mc.mc.add_variable(var1)
var2   = trick.MonteVarFile("simobj.myValue[1]", "Modified_data/monte_carlo/my_mcvar_file.txt", 2)
trick_mc.mc.add_variable(var2)
...
var190 = trick.MonteVarFile("simobj.myValue[190]", "Modified_data/monte_carlo/my_mcvar_file.txt", 190)
trick_mc.mc.add_variable(var190)
```

If this were our only variable, a single monte carlo input data file results in 190 open file handles to the same file at input-processor time.  No wonder we hit our 1024 limit so easily!

This design seems like it could be refactored such that the Trick monte-carlo framework isn't holding on to a file handle per column of the file.  A simple fix might be to change when the file gets opened, so that the file is opened, read, and closed all in one shot.  Or perhaps the architecture could change such that the user only has to instantiate one `MonteVarFile` per file (instead of per-column per-file).

CTLaChance:
There are two solutions that have been discussed regarding this issue.

### Tell and Seek
[...]

PR #498, opened 2017-10-27
Implemented a fix for multiple concurrent file handles.

CTLaChance:
Uses seekg() and tellg() to save and restore the position of the file stream.

This pull request is in response to issue #432.

### DIFF
--- a/trick_source/sim_services/MonteCarlo/MonteVarFile.cpp
+++ b/trick_source/sim_services/MonteCarlo/MonteVarFile.cpp
@@ -9,9 +9,9 @@
 #include "trick/exec_proto.h"
 
 Trick::MonteVarFile::MonteVarFile(std::string in_name, std::string in_file_name, unsigned int in_column, std::string in_unit) : input_file_stream(NULL) {
-    this->name = in_name;
-    this->column = in_column;
-    this->unit = in_unit;
+    name = in_name;
+    column = in_column;
+    unit = in_unit;
 
     set_file_name(in_file_name);
 }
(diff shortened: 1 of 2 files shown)


trick_source/sim_services/MonteCarlo/MonteVarFile.cpp
```diff
@@ -34,6 +34,17 @@ std::string Trick::MonteVarFile::describe_variable() {
 }
 
 std::string Trick::MonteVarFile::get_next_value() {
+
+    // Reopen the file if not open.
+    if(!input_file_stream->is_open()) {
+        input_file_stream->open(this->file_name.c_str(), std::ifstream::in);
+
+        // If the stream position has been set, re-open file to this position.
+        if(this->stream_position != NULL) {
```
alexlin0: This comparison should be made against 0 instead of NULL.  Although NULL will most likely work here, NULL is usually used when comparing pointer values.  The stream_position is most likely a large integer type.

CTLaChance: I see. It's different in Java and C#. I'll make the change real quick.

dbankieris: This check is not necessary. `stream_position` should be initialized to `0` in the constructor. Thereafter, it is set only via `tellg()` in this function. It is always valid.

trick_source/sim_services/MonteCarlo/MonteVarFile.cpp
```diff
@@ -34,6 +34,17 @@ std::string Trick::MonteVarFile::describe_variable() {
 }
 
 std::string Trick::MonteVarFile::get_next_value() {
+
+    // Reopen the file if not open.
+    if(!input_file_stream->is_open()) {
```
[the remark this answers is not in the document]
CTLaChance: The check was intentionally redundant. I don't mind removing it.

trick_source/sim_services/MonteCarlo/MonteVarFile.cpp
CTLaChance: Correct. Another redundant check. I can remove it.

CTLaChance:
Wait a second, getting an error. Will update in a second. Ignore the review request.

Issue #499, opened 2017-10-30
MonteCarlo unit test reporting error.


alexlin0:
Monte Carlo unit test reporting error.  There is a test expecting a return of "1.0000" now instead of "1".

CTLaChance:
I'll fix this real quick.

PR #500, opened 2017-11-01
Modified unit tests to account for new logic.

CTLaChance:
A change I had made to how the get_next_value() function returns values invalidated this unit test. This error stemmed from the changes I made in c8f916e8182dc9b4d9c13b187442e295e67c9290 that allowed non numerical values to be used in the data file. 

Originally, trailing 0s were being stripped before the value was returned.
Now, the value present in the data file is the exact value returned by get_next_value().

I have modified the unit test to reflect this new behavior.


CTLaChance:
This pull request is in reference to issue #499.

Issue #504, opened 2017-11-02
Compiling unit tests returns numerous unused parameter and deprecated conversion warnings.


CTLaChance:
They're just warnings, so it's not breaking compilation, but I want to fix them regardless.

dbankieris:
All warnings must be eliminated! If you aren't sated after fixing unit test warnings, turn on all the `-W` compiler options, build Trick, and take care of those as well.

CTLaChance:
Heck ya!

CTLaChance:
I've merged a few small changes into the master, but I'm not 100% finished with warnings just yet. This merge removed many unused variable warnings, but not all of them. I will leave this issue open until all warnings have been addressed.

PR #515, opened 2017-11-16
Some unit test warnings addressed.

CTLaChance:
This pull request has taken care of all unit test warnings as well as a few compilation warnings.

The issue referenced by this pull request is: #504.

### DIFF
--- a/trick_source/sim_services/JSONVariableServer/JSONVariableServer.cpp
+++ b/trick_source/sim_services/JSONVariableServer/JSONVariableServer.cpp
@@ -24,12 +24,8 @@ Trick::JSONVariableServer::JSONVariableServer() :
 }
 
 Trick::JSONVariableServer::~JSONVariableServer() {
-    if ( listen_dev.hostname != NULL ) {
-        free((char*)listen_dev.hostname) ;
-    }
-    if ( listen_dev.error_handler != NULL ) {
+        free(listen_dev.hostname) ;
         free(listen_dev.error_handler) ;
-    }
 }
 
 const char * Trick::JSONVariableServer::get_hostname() {
--- a/trick_source/sim_services/JSONVariableServer/JSONVariableServer.cpp
+++ b/trick_source/sim_services/JSONVariableServer/JSONVariableServer.cpp
@@ -25,7 +25,7 @@ Trick::JSONVariableServer::JSONVariableServer() :
 
 Trick::JSONVariableServer::~JSONVariableServer() {
     if ( listen_dev.hostname != NULL ) {
-        free(listen_dev.hostname) ;
+        free((char*)listen_dev.hostname) ;
     }
     if ( listen_dev.error_handler != NULL ) {
         free(listen_dev.error_handler) ;
--- a/trick_source/sim_services/VariableServer/VariableServerListenThread.cpp
+++ b/trick_source/sim_services/VariableServer/VariableServerListenThread.cpp
@@ -26,12 +26,8 @@ Trick::VariableServerListenThread::VariableServerListenThread() :
 }
 
 Trick::VariableServerListenThread::~VariableServerListenThread() {
-    if ( listen_dev.hostname != NULL ) {
-        free((char*)listen_dev.hostname) ;
-    }
-    if ( listen_dev.error_handler != NULL ) {
+        free(listen_dev.hostname) ;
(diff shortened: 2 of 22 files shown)


trick_source/sim_services/JSONVariableServer/JSONVariableServer.cpp
```diff
@@ -25,7 +25,7 @@ Trick::JSONVariableServer::JSONVariableServer() :
 
 Trick::JSONVariableServer::~JSONVariableServer() {
     if ( listen_dev.hostname != NULL ) {
-        free(listen_dev.hostname) ;
+        free((char*)listen_dev.hostname) ;
```
dbankieris: Why is this necessary? `listen_dev.hostname` is already a `char*`.

trick_source/sim_services/VariableServer/VariableServerListenThread.cpp
```diff
@@ -27,7 +27,7 @@ Trick::VariableServerListenThread::VariableServerListenThread() :
 
 Trick::VariableServerListenThread::~VariableServerListenThread() {
     if ( listen_dev.hostname != NULL ) {
```
dbankieris: Might as well fix this while you're in here. No need to check for null pointers before freeing.
https://stackoverflow.com/a/6084233

trick_source/trick_utils/comm/test/TCConnectTest.cpp
```diff
@@ -20,7 +20,7 @@ class TCConnectTest : public testing::Test {
          /* device */
          device = (TCDevice *) malloc(sizeof(TCDevice));
          memset( (void *)device,'\0',sizeof(TCDevice) );
-         device->hostname = "127.0.0.1";
+         device->hostname = (char*)"127.0.0.1";
```
dbankieris: This is [not safe](https://arduino.stackexchange.com/a/13443). We should really make `hostname` a `const char*` in `TCDevice`. Barring that, we should use `strdup` here to make a truly mutable string.

trick_source/trick_utils/comm/test/TCConnectTest.cpp
```diff
@@ -20,7 +20,7 @@ class TCConnectTest : public testing::Test {
          /* device */
          device = (TCDevice *) malloc(sizeof(TCDevice));
          memset( (void *)device,'\0',sizeof(TCDevice) );
```
dbankieris: Another opportunity to fix someone else's code! The argument to `sizeof` should be the [variable itself](https://softwareengineering.stackexchange.com/questions/201104/sizeof-style-sizeoftype-or-sizeof-variable), when possible, not its type. That way, if the type changes in the future, this line doesn't need to be changed.

trick_source/sim_services/JSONVariableServer/JSONVariableServer.cpp
CTLaChance: The free or the cast to a char*?

trick_source/sim_services/VariableServer/VariableServerListenThread.cpp
CTLaChance: Got it.

trick_source/trick_utils/comm/test/TCConnectTest.cpp
CTLaChance: Okey poke. Will fix.

trick_source/sim_services/JSONVariableServer/JSONVariableServer.cpp
CTLaChance: Good catch. I have no idea why I initially did that. I'll fix it.

CTLaChance: I remember now! I did that because I changed listen_dev.hostname to a const char* and it was throwing an error when I tried to free a constant.

trick_source/trick_utils/comm/test/TCConnectTest.cpp
CTLaChance: Hostname unfortunately can't be a const. I've tried it and const is discarded down the line when hostname is reassigned. I'll try using strdup and duplicate the literal into the char *.

Issue #495, opened 2017-10-25
Make install fails to successfully install Trick.


CTLaChance:
Entering **sudo make install** causes the installation to fail with the following return statement:

```
cp -r bin include lib libexec share /usr/local
cp: cannot stat 'lib': No such file or directory
Makefile:346: recipe for target 'install' failed
make: *** [install] Error 1
```

Make install is trying to recursively copy the 'lib' directory to /usr/local when it doesn't exist. Adding an empty folder titled 'lib' fixes this issue.

CTLaChance:
You cannot run **make install** without first compiling Trick via **make.**

This is a simple user error. My mistake.

alexlin0:
Reopening to add an error message to the install rule stating that Trick must be compiled first.

PR #516, opened 2017-11-17
Added warning to 'make install'.

CTLaChance:
Hurray!  A warning has been added to the "make install" target that warns users if they haven't compiled Trick first.

This pull request is in reference to issue #495.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -313,7 +313,9 @@ clean_unit_test:
 	    $(MAKE) -C $$i clean ; \
 	done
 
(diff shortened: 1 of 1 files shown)


dbankieris:
I think it would preferable to have `install` depend on `all`, such that Trick is automatically built if necessary. Instead of telling the user what to do, do it for them!

CTLaChance:
In this instance, I like your idea more @dbankieris. Make install should automatically run make for you.

Issue #459, opened 2017-07-24
Monte Carlo fails to run the last set of values from a MonteVarFile input file if the file ends on the same line as the last set of values.


CTLaChance:
## Example

**Success:**
```
 1   # this is a comment line
 2   
 3   0 1.0000 1.5000 2.0000 2.5000
 4   1 1.5000 2.0000 2.5000 3.0000 
 5   2 2.0000 2.5000 3.0000 3.5000
 6   3 2.5000 3.0000 3.5000 4.0000
 7   4 3.0000 3.5000 4.0000 4.5000
 8   5 3.5000 4.0000 4.5000 5.0000
 9   6 4.0000 4.5000 5.0000 5.5000
10   7 4.5000 5.0000 5.5000 6.0000
11  
```
Monte Carlo runs all 8 runs.

**Fail:**
```
 1   # this is a comment line
 2   
 3   0 1.0000 1.5000 2.0000 2.5000
 4   1 1.5000 2.0000 2.5000 3.0000 
 5   2 2.0000 2.5000 3.0000 3.5000
 6   3 2.5000 3.0000 3.5000 4.0000
 7   4 3.0000 3.5000 4.0000 4.5000
 8   5 3.5000 4.0000 4.5000 5.0000
 9   6 4.0000 4.5000 5.0000 5.5000
10   7 4.5000 5.0000 5.5000 6.0000
```
Monte Carlo runs the first 7 runs and terminates before running the 8th.

CTLaChance:
When MonteVarFile's **get_next_value()** function is called, it gets the line from the file and terminates if it detects that the EOF bit was [set.](https://github.com/nasa/trick/blob/master/trick_source/sim_services/MonteCarlo/MonteVarFile.cpp#L35-L38)

When the last line of the input file is received into the buffer, the code detects that the EOF bit was set and it terminates prematurely.

nkapliev:
Should it be closed because PR([#531](https://github.com/nasa/trick/pull/531)) with a fix has been merged?

dbankieris:
Yep!

PR #531, opened 2017-12-12
 Fix MonteCarlo varfile eof line problem. #459

nkapliev:
This pull request is in reference to issue #459.


CTLaChance:
@nkapliev Thank's so much for contributing! We really appreciate it!

nkapliev:
Thank you 😋  
`Trick` is awesome! And documentation is really nice.

PR #556, opened 2018-02-02
Added needed module compiler flags and changed MtvView Vector<Object>…

spfennell:
… to Vector<Vector> for Java 9 DefaultTableModel constructor compatibility. Java directory now compiles under openjdk-9.0.4


dbankieris:
Make sure this still compiles in Java 8, specifically the `--add-modules` flag.

PR #572, opened 2018-03-01
Java 9 Compatibility

spfennell:
Two changes:

Added "--add-modules java.se.ee" compiler flag to javac when version is Java 9 in java/makefile.

Change table_row Vector type from Vector\<Object\> to Vector\<Vector\> for new constructor requirement in DefaultTableModel.


trick_source/java/makefile
```diff
@@ -5,6 +5,13 @@ include ${TRICK_HOME}/share/trick/makefiles/Makefile.common
 
 JAVAC ?= javac
 JAVAC_FLAGS = -g -Xlint:unchecked -Xlint:deprecation
+# delineate 'javac -version' for version number 
+JAVAC_VERSION := $(shell (${JAVAC} -version |& awk '{print $$2}') | awk -F ""."" '{print $$1}')
+
+# include java.se.ee for JAXB xml annotations (for trick-tv) in java version 9
+ifeq ($(JAVAC_VERSION),9)
```
dbankieris: This should probably check for 9 or greater, or we'll be broken again in 10, right?

dbankieris: Try this `javac -version | & awk '{print $2}' | & awk -F'[^0-9]' '{print $1'}`

alexlin0: The "|&" pipe doesn't work on the mac.  The above expression also only returns "1" for all java versions 1.8 and below.  Here is my version using the correct utility, perl.
```
javac -version 2>&1 | perl -ne 'print /(\d+\.\d+)/'
```

alexlin0: My version of the command above returned 2 digits, 9.0 or 1.8.  Here is a version that only returns one.
```
javac -version 2>&1 | perl -ne '/(\d+)\.(\d+)/ ; print +($1 == 1) ? $2 : $1'
```

alexlin0: OK, @dbankieris tells me returning "1" for anything less than 9 is ok.  And not all versions follow the x.x version numbering.  So one more version that only gets the first digit(s)
```
javac -version 2>&1 | perl -ne 'print /(\d+)/'
```

Issue #492, opened 2017-10-17
configure script doesn't check for curl


dbankieris:
We use curl to download the Java JARs we depend on.

PR #576, opened 2018-03-09
Configure script now checks for curl. #492

pappyvw:
```
$ ./configure
...
checking lex library... none needed
checking whether yytext is a pointer... no
checking for bison... /usr/bin/bison
checking for curl... /usr/bin/curl
checking for perl... /usr/bin/perl
checking for python... /usr/bin/python
checking for swig... /usr/local/bin/swig
...
```

### DIFF
--- a/configure
+++ b/configure
@@ -671,6 +671,7 @@ GNUPLOT
 SWIG
 PYTHON
 PERL
+CURL
 BISON
 LEXLIB
 LEX_OUTPUT_ROOT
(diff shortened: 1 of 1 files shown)


alexlin0:
I can't merge this request.  The configure script is generated using autoconf.  These changes will be lost the next time we run autoconf.  To check for curl code needs to be added to autoconf/configure.ac and then we can generate a new configure script.

pappyvw:
I'll close these pull requests for now and reopen them after I get a fix with configure.ac working.

PR #577, opened 2018-03-09
Removing a stray line in configure script.

pappyvw:
Fixing this error in the configure script:

```
$ ./configure
...
checking for g++... g++
checking whether we are using the GNU C++ compiler... yes
checking whether g++ accepts -g... yes
./configure: line 3491: AX_GCC_VERSION: command not found
checking for tee... /usr/bin/tee
checking for ld... /usr/bin/ld
checking for flex... flex
...
```


alexlin0:
I did not realize that AX_GCC_VERSION was no longer working.  AX_GCC_VERSION was a macro provided by autoconf, but is no longer available.  We do need to check for the gcc version, as we need at least gcc 4.8 to compile.   This stackoverflow post shows an alternative that we could try, https://stackoverflow.com/questions/7067385/find-the-gcc-version.  This replacement macro would need to be added to autoconf/configure.ac and then a new configure script would need to be generated.

pappyvw:
I'll close these pull requests for now and reopen them after I get a fix with configure.ac working.

PR #606, opened 2018-04-21
Configure fixes

pappyvw:
This fixes 3 known issues with the configure/configure.ac scripts. For the version checks, I went with the given versions in the Install Guide.

I'm not sure why the order of PYTHON_EXTRA_LIBS and PYTHON_EXTRA_LDFLAGS swapped. Anyone know a way to avoid that?


alexlin0:
Hey @pappyvw ,  Thanks for the fixes!

Issue #588, opened 2018-03-26
Frame Logging data is not recorded for SCHEDULED threads


alexlin0:
Threads that are not AMF or ASYNCHRONOUS never get their data written out to disk.  this is because the top_of_frame and end_of_frame jobs never get run for these threads.

PR #615, opened 2018-05-14
#588 top_of_frame and _end_of_frame fix for scheduled threads

spfennell:
Thank you Steven Scheinholtz for this fix! @scheiny


scheiny:
You’re very welcome!

Issue #614, opened 2018-05-11
LQUEUE usage


abrogley:
Is LQUEUE still supported? I'm updating our old Trick 13 sims to Trick 17.3 and am hitting an issue with structs and the memory manager.

If I have a class:

```c++
/*
PURPOSE:
    (Lqueue demonstration)
*/

#include "trick/include/trick/lqueue.h"

class MyClass {

 private:
   LQUEUE *lq; /* -- queue object */

 public:
   MyClass(){};
   virtual ~MyClass(){};
} ;
```

and an S_define:

```c++
#include "sim_objects/default_trick_sys.sm"

##include "MyClass.hh"

class MySimObject: public Trick::SimObject {

   public:
      MyClass classObject;
      MySimObject(){}

} ;

MySimObject simobj;
```

This will build with the following memory manager warning:
```
./S_main_Linux_5.4_x86_64.exe sie
MemoryManager:WARNING:(build/path/to/io_MyClass.cpp:46): Couldn't find an io_src_sizeof_ function for type _LQUEUE[io_src_sizeof__LQUEUE()].

MemoryManager:WARNING:ATTRIBUTES for type "_LQUEUE" not found.
```

And if I hide the LQUEUE object from the ICG I get the standard warning about using TRICK_ICG in header files.

What's the preferred way to use LQUEUE?

PR #621, opened 2018-05-16
Add lqueue.h lstack.h, and dllist.h to the list of Trick files that get ICGed

jmpenn:
This resolves Issue #614.


abrogley:
Tried this out and it fixes my issues. Thanks John!

Issue #603, opened 2018-04-18
Add ability to print out environment and makefile variables


alexlin0:
If we add these lines to the top level sim makefile

```makefile
print-%:
        @echo '$*=$($*)'
```
This will let you print environment variables and makefile variables from the command line

e.g.

```
make print-TRICK_SFLAGS
make print-TRICK_CFLAGS
make print-TRICK_SYSTEM_CFLAGS
```

iamthad:
Isn't this already in [Makefile.sim](https://github.com/nasa/trick/blob/master/share/trick/makefiles/Makefile.sim#L79)?

alexlin0:
I thought I already put it in there.  I couldn't find it the other day.  I must have been in a different version... Thanks for point it out.  Yup, closing this issue.

iamthad:
I just tried this on one of my sims and it didn't seem to work. Can you confirm that it does?

alexlin0:
It should be deleted.

spfennell:
Fixed in #623

PR #623, opened 2018-05-22
#603 restored 'make print-%' in sim directory

spfennell:


bin/trick-CP
```diff
@@ -135,14 +136,20 @@ convert_swig: build/S_library_swig
 S_define_exp:
 	$(TRICK_CC) -E -C -xc++ ${TRICK_SFLAGS} $(TRICK_SYSTEM_SFLAGS) S_define > $@
 
+# prints the value of a makefile variable, example invocation "make print-TRICK_CXXFLAGS"
+# This rule is used by trick-config
```
alexlin0: Reading this comment reminded me that bin/trick-config uses Makefile.sim,  I would suggest creating a small makefile in the same directory that Makefile.sim lives in that includes Makefile.common and has one rule for "print-%:"

Issue #627, opened 2018-06-05
Sims that have source files in a directory containing a `src` directory don't compile


dbankieris:
`make_makefile_src` [specifically looks](https://github.com/nasa/trick/blob/1015f6ec9001970504d215801babf9c00d4b0bc7/libexec/trick/make_makefile_src#L154) for `src` in each file's path and sticks it in `$files_by_dir{$sp_dir}{src_dir}`, which is [then used](https://github.com/nasa/trick/blob/1015f6ec9001970504d215801babf9c00d4b0bc7/libexec/trick/make_makefile_src#L268) when listing the objects files in `Makefile_src`. If you have source files in `foo` and `foo/src`, both entries in `Makefile_src` have `src` in their path, which is wrong. Stop giving special treatment to `src`!

PR #629, opened 2018-06-07
Fix "objects:" in makefile_overrides

### LINKED ISSUE #626: "objects:" doesn't work in makefile_overrides
The per-extension `*_objects:` keywords don't work either.

dbankieris:
Fixes #626 & #627

### DIFF
--- a/libexec/trick/make_makefile_src
+++ b/libexec/trick/make_makefile_src
@@ -199,13 +199,6 @@ foreach $k ( sort keys %files_by_dir ) {
     }
 }
 
-# set the "dir_num" of each directory.
-foreach $k ( sort keys %files_by_dir ) {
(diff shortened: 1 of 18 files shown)


jmpenn:
This isn't directly related to the bug your fixing but a minor thing I noticed in the same file.
In  the following line:

```elsif ( s/([cfhy]|C|cc|cxx|cpp|CPLUSPLUS)_objects\s*:\s*// ) {```

 In ```[cfhy]```, shouldn't it be ‘l’ for lex, rather than 'h'?

You think I should make a new issue or you want to sneak this in?

dbankieris:
I see `[cfhy]` in the old version on line 383, but it's not in the new version.

PR #631, opened 2018-06-14
Add M_SIGN macro

pappyvw:
The standard math.h library does not contain a sign function. Adding
a simple macro to trick to output (-1, 0, 1) based on the sign of the
input expression. This can evaluate sign for C++ signed types.

```
    void test_M_SIGN() {
        printf("50.0   %d\n", M_SIGN(50.0));
        printf("-13    %d\n", M_SIGN(-13));
        printf("0      %d\n", M_SIGN(0));
        printf("0.0    %d\n", M_SIGN(0.0));
        printf("-0     %d\n", M_SIGN(-0));
        printf("-0.0   %d\n", M_SIGN(-0.0));
        printf("inf    %d\n", M_SIGN(std::numeric_limits<double>::infinity()));
        printf("-inf   %d\n", M_SIGN(-std::numeric_limits<double>::infinity()));
    }
```

```
50.0   1
-13    -1
0      0
0.0    0
-0     0
-0.0   0
inf    1
-inf   -1
```

### DIFF
--- a/include/trick/trick_math.h
+++ b/include/trick/trick_math.h
@@ -11,6 +11,10 @@
 #define M_ABS(x) ((x) < 0 ? -(x) : (x))
 #endif
 
+#ifndef M_SIGN
+#define M_SIGN(x) (((x) > 0) - ((x) < 0))
+#endif
+
 #if ( __ghs )
 #ifndef M_PI
 #define M_PI     3.141592653589793238460


iamthad:
Any reason to not just use [`copysign`](http://en.cppreference.com/w/c/numeric/math/copysign)?

```c
copysign(1.0,  x)
```

would do the same as your macro.

pappyvw:
You know, I initially thought copysign would not work for my needs but yes this works for me. Thanks!

lordmundi:
it's faster too.  I was curious so timed copysign versus a couple of
algorithms to implement M_SIGN (I don't think a standard one exists).  I
got:

Time taken by copysign function for 10000000 iterations: 41503 microseconds
Time taken by M_SIGN implementation for 10000000 iterations: 55370
microseconds

I even tried some optimized algorithm from here:
http://graphics.stanford.edu/~seander/bithacks.html#CopyIntegerSign

and that was still slower than copysign.  I'm guessing they have some
optimizations for CPU instructions in there. anyhow... copysign appears the
better way to go regardless unless someone can find a faster way!

FG

PR #636, opened 2018-06-21
Print a stack trace when exec_terminate is called

### LINKED ISSUE #635: Print a stack trace when exec_terminate is called
So you can point the finger at the culprit!

dbankieris:
Closes #635

### DIFF
--- a/include/trick/Executive.hh
+++ b/include/trick/Executive.hh
@@ -283,6 +283,17 @@ namespace Trick {
 
             ~Executive() ;
 
+            /**
+             * Print a demangled stack backtrace of the caller function to FILE* out.
+             *
+             * @param out the file to print to
+             * @param max_frames the maximum numbers of frames to trace
+             */
+            static void print_stack_trace(FILE *out = stderr, unsigned int max_frames = 63);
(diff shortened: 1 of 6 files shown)


iamthad:
Any reason to use this new backtrace code rather than just shelling out to `gdb` like the signal handlers do?

dbankieris:
`exec_terminate` is the normal path by which a sim ends, so we don't want to always invoke gdb. Although you could break on the function yourself in gdb, that wasn't an option for the requester of this feature. He wanted an alternative method to identify the called of `exec_terminate`.

alexlin0:
I don't think this is a capability we need.   The user should send exec_terminate a useful message indicating the reason for termination rather than asking Trick to provide a stack trace on exit.  

The old backwards compatible <code>exec_terminate(const char *file_name, const char *error) </code> doesn't include the line number where the call is made, the newer <code>exec_terminate_with_return(int ret_code, const char *file_name, int line, const char *error)</code> provides a field for the line number.

dbankieris:
Only squeaky wheels get attention around here.

Issue #618, opened 2018-05-16
Add TRICK_SWIG_FLAGS


dbankieris:
Alas, users have no way of adding flags to the invocation of SWIG! Change `SWIG_CFLAGS` to `TRICK_SYSTEM_SWIG_FLAGS` (to be used by Trick only), and add `TRICK_SWIG_FLAGS` as the user-facing interface.

PR #640, opened 2018-06-27
#618 add TRICK_SWIG_FLAGS, TRICK_SYSTEM_SWIG_FLAGS, TRICK_SWIG_CFLAGS, TRICK_SYSTEM_SWIG_CFLAGS...

spfennell:
…
\*\_FLAGS are swig options, \*\_CFLAGS are g++ compiler options for swig objects. TRICK_SYSTEM\_\* options should not be changed by trick users.

### DIFF
--- a/libexec/trick/make_makefile_swig
+++ b/libexec/trick/make_makefile_swig
@@ -231,8 +231,8 @@ SWIG_SRC = \$(subst .i,.cpp,\$(SWIG_I)) $swig_src_dir/top.cpp
 
 \$(SWIG_SRC) : %.cpp: %.i | \$(SWIG_I)
 \t\$(PRINT_SWIG)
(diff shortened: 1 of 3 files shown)


dbankieris:
Hmm, issue #618 may have been created in error. Trick doesn't currently set `SWIG_FLAGS` anywhere, so it already is a user-facing variable, and is referenced as such in the [wiki](https://github.com/nasa/trick/search?q=SWIG_FLAGS&unscoped_q=SWIG_FLAGS&type=Wikis). Trick _does_ set `SWIG_CFLAGS`, and users shouldn't override it (although appending would be ok), but I doubt they'd ever need to do so. Perhaps this whole pull request is a mistake! Need to consult with @alexlin0 when he returns.

alexlin0:
For consistency I like the changes this pull request makes.  Flags prepended with "TRICK_SYSTEM" are set by Trick and should not be changed by the user.  Flags only prepended with "TRICK" like are user facing, start empty, and are for the user to change.

dbankieris:
Also, the [wiki](https://github.com/nasa/trick/wiki/Trickified-Project-Libraries) and [trickify.mk](https://github.com/nasa/trick/blob/bb9b136a66dba741140160e586cf87fe6b80f15f/share/trick/makefiles/trickify.mk) need to be updated.

spfennell:
@dbankieris please review changes. Also check out the [Environment Variables](https://github.com/nasa/trick/wiki/Environment-Variables) wiki page.

PR #645, opened 2018-07-11
add --with-swig=<path to swig> option to configure

spfennell:


spfennell:
@dbankieris I hear you have a test case

configure
```diff
@@ -1385,6 +1387,7 @@ Optional Features:
 Optional Packages:
   --with-PACKAGE[=ARG]    use PACKAGE [ARG=yes]
   --without-PACKAGE       do not use PACKAGE (same as --with-PACKAGE=no)
+  --with-swig[=DIR]       SWIG root directory
```
dbankieris: One of my frustrations with `configure` has been the vagueness of option descriptions. I would prefer something less ambiguous here, such as `directory containing the swig executable`.

spfennell: I agree. I was just trying to conform! Maybe I'll update some of the other help strings.

dbankieris: That would be great!

PR #658, opened 2018-08-16
Warning fixes

jmpenn:
Bison and Flex both generate warnings from generated code. The code is fine.

void * Trick::MessageThreadedCout::thread_body() needs to return a value, specifically NULL;


jmpenn:
BTW: I tested these on Linux and Mac OS X.

PR #662, opened 2018-08-21
Trick View: use fixed width for binary format

### LINKED ISSUE #661: Trick View: used fixed width for binary format
Binary format should pad with leading zeros so the number of displayed digits is constant.

dbankieris:
Approach stolen from http://javadevnotes.com/java-integer-to-fixed-length-string

Closes #661


dbankieris:
If we raise our requirement to Java 11, I can use [String::repeat](https://download.java.net/java/early_access/jdk11/docs/api/java.base/java/lang/String.html#repeat(int)) :smile:

PR #666, opened 2018-09-12
Fix incorrect initialization of units in trick-qp.

jmpenn:
Trick-qp initialized units elements to "--" which is incorrect.
The <units> element optionally specifies units to which the recorded should be converted.
When non-null a <units> specification should be generated. When null the <units> element should not be generated. The bogus <units> -- <\units> was causing the fermi tape generation tool to crash because it was attempting to convert "m/s2" etc. to "--".


alexlin0:
I should reject this just because of the issue number.

jmpenn:
It'll be fine! :smiling_imp:

Issue #667, opened 2018-09-12
Deliberate divide by zero craziness in "deuler"  trick math utils


jmpenn:
What were they thinking!!? :scream:

Each of the c files below contains the following lines: 

```
double zero = 0.0;
angle[0] = angle[1] = angle[2] = 0.0 / zero;
```

deuler_123.c
deuler_123_quat.c
deuler_132.c
deuler_132_quat.c
deuler_213.c
deuler_213_quat.c
deuler_231.c
deuler_231_quat.c
deuler_312.c
deuler_312_quat.c
deuler_321.c
deuler_321_quat.c

lordmundi:
are these in an error condition where they are trying to force the numbers
to be "NaN" ?

FG

iamthad:
It does appear to be their way of obtaining a NaN in an error case. For example, https://github.com/nasa/trick/blob/master/trick_source/trick_utils/math/src/deuler_123.c#L120

jmpenn:
math.h defines NAN, and isnan()

If that's what they are trying to do,I think I'll use NAN from math.h instead.

jmpenn:
Given the message to which you referred, and the return code   ```ret = TM_ANG_NAN;```
, I agree with you.

I'll the following fix:

Delete: ```double zero = 0.0;```

and

Replace: ```angle[0] = angle[1] = angle[2] = 0.0 / zero;```
With: ```angle[0] = angle[1] = angle[2] = NAN;```

spfennell:
fixed in pull request #668

lordmundi:
you guys may want to double check portability... i'm guessing they did this
originally because not all implementations define NAN.  From here:
http://pubs.opengroup.org/onlinepubs/007904975/basedefs/math.h.html

"This symbolic constant is only defined if the implementation supports
quiet NaNs for the float type."

Now, my guess is that most implementations you guys care about support
this... but if you want to be sure, it sounds like you can add:

#ifdef NAN
/* NAN is supported */
#endif

anyhow.  just adding that info out there.

Happy Halloween!

FG

jmpenn:
We looked at this again, and came to the conclusion that we're happy with it as is. If someone can show us a relatively modern system that doesn't support quiet NaNs, we could could look at it further.
Thanks

PR #668, opened 2018-09-14
Replace unreliable 'divide-by-zero' way of generating NaNs with math.h NAN in *deuler* routines.

jmpenn:
Previous code generated FP div by zero error.
This fix returns NaNs in each of the angle elements.


spfennell:
#667

Issue #669, opened 2018-09-18
command line arguments C interface lacks "void" for zero argument functions


excaliburtb:
i.e.
`const char * command_line_args_get_default_dir() ;`
should be 
`const char * command_line_args_get_default_dir(void) ;`

to avoid compiler warnings.

excaliburtb:
Just FYI, I get the warning with -Wstrict-prototypes compiler flags with gcc v4.8.5

PR #670, opened 2018-09-18
Explicitly specify void arg for C functions that take no arguments.

jmpenn:
In prototypes and definitions. This doesn't apply to C++.

Apparently some compilers emit warning messages in args are not explicitly specified as void.


jmpenn:
See issue #669

Issue #685, opened 2018-10-22
Valgrind/Callgrind exposes threading issue in Trick shutdown


ddj116:
When running a sim through callgrind via `valgrind --tool=callgrind ./S_main... RUN.../input.py`, often the process will hang for hours or even days attempting to shut down.  Although my group is currently seeing this in our Trick 13 version, the core issue appears to still exist in the latest Trick 17 version.  I can replicate this in `SIM_cannon_eulercromer` -- when a nominal valgrind run may take 10 seconds, often it will hang for 30+ min eventually requiring me to kill the process.

When using `callgrind_control -b`, I can see it's stuck somewhere in the shutdown logic.  I see this same backtrace for a long period of time.  It appears to have something to do with the data recording thread, but I'm not exactly sure of the culprit. 

```
~/dev/trick > callgrind_control -b 55162
PID 55162: ./S_main_Linux_4.8_x86_64.exe RUN_test/input.py
sending command status internal to pid 55162

  Frame: Backtrace for Thread 1
   [ 0]  pthread_mutex_trylock (68853711 x)
   [ 1]  Trick::DataRecordDispatcher::shutdown() (1 x)
   [ 2]  DataRecordDispatcherSimObject::call_function(Trick::JobData*) (8 x)
   [ 3]  Trick::JobData::call() (6 x)
   [ 4]  Trick::Executive::shutdown() (1 x)
   [ 5]  master(int, char**) (1 x)
   [ 6]  main (1 x)
   [ 7]  (below main) (1 x)
   [ 8]  _dl_runtime_resolve_xsave (1 x)
   [ 9]  0x0000000000578010 (1 x)
   [10]  0x0000000000001030

  Frame: Backtrace for Thread 2
   [ 0]  start_thread (3 x)
   [ 1]  clone

  Frame: Backtrace for Thread 3
   [ 0]  start_thread (3 x)
   [ 1]  clone

  Frame: Backtrace for Thread 4
   [ 0]  write (452 x)
   [ 1]  Trick::DRBinary::format_specific_write_data(unsigned int) (452 x)
   [ 2]  Trick::DataRecordGroup::write_data(bool) (5 x)
   [ 3]  Trick::DRDWriterThread::thread_body() (1 x)
   [ 4]  Trick::ThreadBase::thread_helper(void*) (3 x)
   [ 5]  start_thread (3 x)
   [ 6]  clone
```

[...]

PR #699, opened 2018-11-08
Fix lockup when running Valgrind. Ref #685

jmpenn:
This fixes the problem described in Issue 685.


ddj116:
This patch appears to fix the issue as reproduced in `SIM_cannon_euler_cromer` as described in #685.  I'm now testing in a much larger sim where each run takes about 30 min, so I should know by the end of the day if we get the same fix in our larger system.

Thanks!

ddj116:
Larger sim testing is going well so far, once y'all merge I'll patch our side officially with this content.

Issue #702, opened 2018-11-13
MonteCarlo: Correlated / dependent dispersions


garyt2:
Addresses the situation in which certain variables are to be dispersed according to rules that make them functions of other dispersed variables.

Simplest example is to disperse some value and have it assigned directly to *n* other values.  This is currently done by making *n+1* dispersions, each with the identical range and seed so the same random number is generated for every variable -- but it is generated to be identical, not assigned to be identical.

More complex problems are not directly addressable, e.g. z = x * y, where x and y are dispersed.

A solution to this would be to introduce code-injection points into the MC variable dispersion process so that, for example:
```
<disperse x[0]>
for ii in range (1,n+1):
  x[ii] = x[0]
```
or
```
<disperse x>
<disperse y>
z = x * y
```
Or much more elaborate assignments, basically anything that can be handled in a Python script.

monte_input would look like:
```
<disperse x>
<disperse y>
execfile("compute_z.py")
<disperse p>
...
```

garyt2:
I'll submit a pull-request with a potential fix.

Issue #700, opened 2018-11-13
MonteCarlo: Integers recorded as floats


garyt2:
Minor issue; when dispersing integers using MonteVarRandom, a floating-point value is generated which is then cast to an integer upon assignment to the variable.

However, the value recorded in monte_input is the floating-point value, not the integer.

Whether the value is recorded as an integer or its original floating-point value makes no difference to the execution, but it would be easier to read the monte_input if integer values were expressed as integers.

garyt2:
I'll submit a pull-request with a potential fix.

PR #704, opened 2018-11-14
700 Adding integer flag to MonteVarRandom

garyt2:


jmpenn:
My issue with this proposed change is that it creates a discrete distribution from a continuous one, thus altering its fundamental characteristics in a not-commonly-understood way.

We do have a mechanism, called *MonteVarCalculated* that one could use to generate values, distributed in any way they need. Trick::MonteVarCalculated::get_next_value() returns values that are calculated in a *monte_master_pre* jobs,  executed right before a monte-run is dispatched to a monte-slave.

iamthad:
Another consideration: casting to `int` will truncate the value, rather than rounding it.

garyt2:
> My issue with this proposed change is that it creates a discrete distribution from a continuous one, thus altering its fundamental characteristics in a not-commonly-understood way.

I agree with this sentiment, but the reality is that users are already generating discrete distributions from MonteVarRandom.  Users are assigning values generated from MonteVarRandom to `int` types.  It would be better to templatize MonteVarRandom to tailor the distributions, but that's a fair bit more work, not something I have time for.

The only thing this PR does is record that casting in monte_input so that it is *easier* to understand what is happening.

e.g.
my_model.fail_index = 4.897263985269

vs.
my_model.fail_index = 4

They both do the same thing, but the latter is more instructive to a general user wanting to find the specs for a particular run.

> 
> We do have a mechanism, called _MonteVarCalculated_ that one could use to generate values, distributed in any way they need. Trick::MonteVarCalculated::get_next_value() returns values that are calculated in a _monte_master_pre_ job, executed right before a monte-run is dispatched to a monte-slave.

I think this comment is addressing the MonteVarExecCode modification (#702 / #706)  (correct me if I'm wrong).  I did explore this as a solution but came across 2 difficulties:
- Every variable must be added individually.  Some of our post-processing code that users are injecting in their own way involves dozens of variables.  It's not a big deal, but I'm going to face a not-insignificant amount of griping and heel-dragging if I ask for all variables to be added as MonteVarCalculated instances instead.
- More significantly, I believe the code processed through monte-var-pre jobs has to be compiled code.  That's just not going to work for autocoded models; developers *will not* add dispersion code to their models.  If they could/would, these asssignments could be init-class jobs anyway and the problem would not even exist.

garyt2:
> Another consideration: casting to `int` will truncate the value, rather than rounding it.

Definitely a concern.  And I basically flipped a coin over which way to go with this.  The simple way -- the one I settled on -- is to leave the value unchanged from the natural casting that occurs when the random value is assigned to an int.  This means that calling `cast_to_integer` as part of the variable configuration has no effect other than recording the assigned value as an integer in `monte_input`.

The risk is the situation when the random number generator hits the upper bound (I'm not sure whether this is possible ... how does the random number generator work?).  E.g. generating a random number between 0 and 2 would truncate to either 0 or 1 with equal probability ... unless it hits 2.00000000000

It would be more intuitive for that configuration to generate 0,1,or 2 with equal probability, but that is not the way users are currently configuring MonteVarRandom to generate random integers.

alexlin0:
I don't like this change either.  If truncating random values isn't the right way to do this, then I don't want to be the one that owns this code.

If this is the code you want to use, an alternative solution is to create a new class that inherits from MonteVarRandom and modifies the output values as you want to.  This can be done in a user model area rather than having to be part of Trick core.

garyt2:
> I don't like this change either. If truncating random values isn't the right way to do this, then I don't want to be the one that owns this code.

Whether truncating random values is right or wrong is moot, it is currently the only way (that I know of) to use the MC process to generate random integers.  If that is not intended to be used, the Trick documentation needs to be very clear that MC dispersion of integers is not currently supported (or provide a mechanism that does support dispersion).  If that is the intended mechanism for generating integers, then this change should have no effect.

Nevertheless, I understand that you do not want to appear to support a technique that has potential issues, so I will withdraw this PR.

Could you offer any suggestions as to how I should direct users to disperse random integers?

> 
> If this is the code you want to use, an alternative solution is to create a new class that inherits from MonteVarRandom and modifies the output values as you want to. This can be done in a user model area rather than having to be part of Trick core.

This is troubling.  Are you really advocating that individual NASA projects should be making modifications to Trick for their own use?

dbankieris:
@jmpenn @alexlin0 I would encourage you to read issue #700 and comment on whether or not you think it's valid. Then we can discuss if *any* solution is needed before debating a particular implementation (such as this pull request).

dbankieris:
@garyt2 For flat distributions, a random number generator's range is usually [min, max), where min is included but max is excluded, so truncation is the appropriate way to generate evenly-distributed ints.

PR #705, opened 2018-11-14
701 processing mc flag

garyt2:
Adds a usable flag to process input.py instructions independent of the enabled flag

### DIFF
--- a/include/trick/MonteCarlo.hh
+++ b/include/trick/MonteCarlo.hh
@@ -16,7 +16,7 @@
 
 #ifdef SWIG
 // This instructs SWIG to use dynamic_cast and return one of the derived type for the get_variable function.
-%factory(Trick::MonteVar * Trick::MonteCarlo::get_variable, Trick::MonteVarCalculated, Trick::MonteVarFile, Trick::MonteVarFixed, Trick::MonteVarRandom) ;
+%factory(Trick::MonteVar * Trick::MonteCarlo::get_variable, Trick::MonteVarCalculated, Trick::MonteVarFile, Trick::MonteVarFixed, Trick::MonteVarRandom, Trick::MonteVarExecCode) ;
 #endif
 
 namespace Trick {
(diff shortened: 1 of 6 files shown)


include/trick/MonteCarlo.hh
```diff
@@ -449,6 +454,20 @@ namespace Trick {
          */
         bool get_enabled();
 
+        /**
+         * Sets #processing_input_for_MC.
+         *
+         * @see @ref MonteCarloEnabling "Enabling Monte Carlo"
```
garyt2: This has not been updated; need guidance from Trick team on how best to do this.

include/trick/montecarlo_c_intf.h
```diff
@@ -23,6 +23,18 @@ void mc_set_enabled(int enabled);
  */
 int mc_get_enabled(void);
 
+/**
+ * @relates Trick::MonteCarlo
+ * @copydoc MonteCarlo::set_processing_input_for_MC
```
garyt2: I'm not sure what this does, may need to add something to documentation before merging or change this line.

trick_source/sim_services/MonteCarlo/MonteCarlo_funcs.cpp
```diff
@@ -11,12 +11,23 @@
 
 void Trick::MonteCarlo::set_enabled(bool in_enabled) {
     this->enabled = in_enabled;
+    if (in_enabled) {
+      set_processing_input_for_MC( true);
```
shahzadlone: Fix space.
```diff
-      set_processing_input_for_MC( true);
+      set_processing_input_for_MC(true);
```

alexlin0:
I believe that we have the ability to do this without code modification.  I believe that you can test if you are a slave and can test if monte_carlo is enabled, you achieve the desired effect.

```python
if trick.mc_is_slave() and trick.mc_get_enabled():
    print "I am a slave processing input file"
```

garyt2:
I think this has become OBE, so I'll close it.  The accumulated implications of trying to piece together patches to cover the use cases we need made doing so unworkable ... suggesting instead the need for an independent refactor/ comprehensive interface which is now underway.

The driver for this work was to get a mechanism in place to allow input file processing during re-runs to match that during first-runs.  Because `mc_get_enabled()` returns false during a re-run, any logic embedded in `if trick.mc_get_enabled()` will not be processed during a re-run, the slave status is moot.  Consequently the processing of the input files will still not match so no, this does not appear to provide the solution.

PR #706, opened 2018-11-14
702 Adding MonteVarExecCode

garyt2:
Adds a MonteVar* type that provides a code-injection point into the dispersion processing.

### DIFF
--- a/include/trick/MonteVarExecCode.hh
+++ b/include/trick/MonteVarExecCode.hh
@@ -1,0 +1,71 @@
+/*
+  PURPOSE: (Monte carlo structures to inject executable logic based on MC
+            dispersions)
+  ASSUMPTIONS AND LIMITATIONS:
+    (The order in which this type of MonteVar gets added to trick_mc.mc is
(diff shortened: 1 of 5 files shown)


include/trick/MonteVarExecCode.hh
```diff
@@ -0,0 +1,71 @@
...
+         *
+         * @param name the address of the Python file relative to the execution directory
+         */
+        MonteVarExecCode(std::string name);
+
+        // Composite the various elements of this MonteVar.
+        virtual std::string describe_variable();
+
+        protected:
+        virtual std::string get_next_value();
```
shahzadlone: Which classes would override these virtuals?

shahzadlone:
I think you have to resolve a conflict.

garyt2:
This has become OBE, so I'll close it. The accumulated implications of trying to piece together patches to cover the use cases we need made doing so unworkable ... suggesting instead the need for an independent refactor/ comprehensive interface which is now underway.

PR #727, opened 2019-01-25
Vector Optimize to Save on Re-Allocation costs.

shahzadlone:
We know how many more elements we are going to push into ```listLogGroupIterators_``` so use that to our advantange and reserve inadvance to save on re-allocation costs.


shahzadlone:
Is there anything from my side I need to provide more

alexlin0:
Thank you for the contribution

Issue #728, opened 2019-01-28
Shouldn't set TCP_NODELAY socket option for UDP sockets


iamthad:
Currently we attempt to set `TCP_NODELAY` for both UDP [clients](https://github.com/nasa/trick/blob/master/trick_source/trick_utils/comm/src/tc_init_udp_client.c#L65) and [servers](https://github.com/nasa/trick/blob/master/trick_source/trick_utils/comm/src/tc_init_udp_server.c#L65). 

The TCP_NODELAY socket option disables Nagle's algorithm on TCP sockets, improving latency at the cost of throughput. This option is TCP-specific, and it is an error to use it on any other kind of socket, including UDP. It is also unnecessary, since UDP is datagram-based and does not buffer data to be transmitted.

PR #729, opened 2019-01-28
Disable setting TCP_NODELAY on UDP sockets

iamthad:
The TCP_NODELAY socket option disables Nagle's algorithm on TCP sockets,
improving latency at the cost of throughput. This option is
TCP-specific, and it is an error to use it on any other kind of socket,
including UDP. It is also unnecessary, since UDP is datagram-based and
does not buffer data to be transmitted.

See #728


alexlin0:
Thanks @iamthad

Issue #738, opened 2019-02-27
Put Derek's warning in when maximum data record file size reached.


alexlin0:
When you hit the maximum data record file size, default 1Gb, then print a warning that the max was reached and how to fix it.

dbankieris:
And print it at the bottom of the file itself, not in the terminal where no one will ever see it.

iamthad:
What is the rationale for limiting the size by default?

dbankieris:
Alex got tired of people accidentally leaving sims running overnight and filling up our filesystem servers.

spfennell:
@dbankieris It looks like I ended up printing the error to std::cerr, maybe we had an in-person discussion about it and decided to go with that? Or maybe I just ignored your request to print at the end of file. Do you remember?

dbankieris:
I do not. Maybe Alex vetoed me. I still think it belongs in the file.

PR #739, opened 2019-02-27
#738 add Derek's warning

spfennell:


spfennell:
We may want to make edits to the warning message before merging.

trick_source/sim_services/DataRecord/DataRecordGroup.cpp
```diff
@@ -692,6 +693,14 @@ int Trick::DataRecordGroup::write_data(bool must_write) {
             writer_num++ ;
 
         }
+
+        if(!max_size_warning && (total_bytes_written > max_file_size)) {
+            std::cerr << "WARNING: Data record max file size " << (max_file_size>>10) << "KB reached. Contact Derek Bankieris regarding any concerns.\n"
```
dbankieris: Nice try, bucko! Also, kilobytes? Nobody wants to see a number that big! Also also, shifting will result in a value of 0 if they set the limit below 1 KB. Better use real division! Also also also, look at this cool new suggestion feature!
```suggestion
            std::cerr << "WARNING: Data record max file size " << (static_cast<double>(max_file_size)/(1<<20) << "MB reached.\n"
```

PR #759, opened 2019-04-24
Python 3 Compatibility for variable_server.py

nmerlene:
* Convert string to bytes in `variable_server.py` for Python 3
* Method socket.sendall() (called in `VariableServer.send()`) expects bytes to be sent, rather than a string
* Python 3 stores text strings as unicode (not ASCII) by default
* Therefore, strings must be explicitly converted to bytes for Python 3 compatibility
* This works for both Python 2 and Python 3 (tested on Python 2.7, Python 3.6, and Python 3.7)


dbankieris:
Thanks for the contribution!

PR #778, opened 2019-05-22
fix: TPROCTE_HOME Being Used In Place Of TPRO_HOME

esduran:
### Introduction

Hi Trick team,

My name is Esteban Duran. I am a former intern of Frank Delgado's at JSC. I now work as a flight software engineer at Firefly Aerospace. The software team and I found this easy to miss discrepancy when enabling real-time clock support in Trick. Thanks for making such a great framework!

### Bug

We are trying to enable the use of a real-time clock for Trick and have encountered an issue
Per the documentation, adding the _--with-tpro="path/to/driver"_ when executing the configure script should enable TPRO CTE support. What we found was that the configure file would completely ignore the path provided and simply print out:

`configure: error: could not find /h/tpro.h`

It looks like in the autoconf/configure.ac on line 63, there are still references to the TPROCTE_HOME variable and it is being used interchangeably with TPRO_HOME. The same is being done in the trick/configure file on line 7079 where the TPROCTE_HOME variable is being set to the withval variable contents but the TPRO_HOME variable is being used to check for the tpro.h file.
The TPROCTE_HOME variable was never used and the TPRO_HOME variable will always be an empty string. Thus, the configuration file will always check for only /h/tpro.h instead of <user-provided-path-to-tpro>/h/tpro.h.

### Fix

Change all references of the TPROCTE_HOME variable in the configure file to TPRO_HOME.


dbankieris:
Thanks for the fix!

PR #780, opened 2019-05-29
Autogenerate SWIG dependencies

### LINKED ISSUE #505: When available, use SWIG 2.0 to autogenerate dependencies
SWIG 2.0 has all the `-M` options that gcc provides for autogenerating makefile dependency rules. We don't yet require SWIG 2.0, but I don't want to forget about this feature. So this issue will just sit here mocking us until the future arrives.

Revert 2db341cff86b3d25cfc5216811d40e341105a9ba when possible.

Or, if we can check the version of SWIG from the makefile, add the flags conditionally.

dbankieris:
Require SWIG >= 2.0

Closes #505

### DIFF
--- a/autoconf/configure.ac
+++ b/autoconf/configure.ac
@@ -126,8 +126,7 @@ AC_DEFUN([AX_GTEST_HOME],[
 AC_DEFUN([AX_SWIG_BIN],[
     AC_ARG_WITH([swig],
                 AS_HELP_STRING([--with-swig@<:@=DIR@:>@], [path of directory containing the SWIG executable.]),    
-                SWIG_BIN="$withval",
-                SWIG_BIN=/bin:/usr/bin:/usr/local/bin:/sw/bin
+                PATH="$withval:${PATH}"
                )
     AC_SUBST([SWIG_BIN])
 ])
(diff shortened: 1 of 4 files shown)


alexlin0:
This commit does not work on the Mac.  swig installed through Homebrew is found in /usr/local/bin/swig and that is not checked in the configure script where it was before.

spfennell:
> This commit does not work on the Mac in our continuous integration. It cannot find swig. The current location is /usr/local/bin/swig.
> 
> This works for me because I have /usr/local/bin in my path. May have to add /usr/local/bin to PATH in configure script?

@alexlin0 @ddj116 This is causing issues for people who don't want /usr/local/bin in their path. Maybe we could add /usr/local/bin to the END of the path variable instead of the beginning? Or only search /usr/local/bin for SWIG as a last resort? hmm...

ddj116:
> Maybe we could add /usr/local/bin to the END of the path variable instead of the beginning

I believe this would fix our particular use case, but I'll defer to ya'll on what the best solution would be. I've never built a real configure script before so I don't really understand the best conventions for these types of things.

Issue #776, opened 2019-05-15
trick_cpr.cpr.load_default_data takes a long time


alexlin0:
This default data class job can take minutes to finish for large sims.  Why is it taking so long?

Issue #526, opened 2017-12-11
Investigate read_checkpoint_from_string performance in 17 vs 15


dbankieris:
NExSyS's `SIM_NextSTEP_ref` ran into an issue when upgrading to Trick 17. The symptom was that every interaction with the sim from the dcapp caused the sim to briefly hang. Rapid interactions caused the sim to hang for dozes of seconds. Inspecting the situation with gdb revealed:

1. The main thread was blocked on event processing, waiting for the Python parser to become available.
2. A variable server thread had the lock on the Python parser and was executing `read_checkpoint_from_string`, which is what the dcapp uses to set variables.

The variable server thread's stack trace had a dozen or so calls related to restoring STLs. I believe Trick 17 is more aggressive about supporting STLs than 15, so that might be the source of the performance difference. Modifying the dcapp to use the more modern `var_set` resolved the issue, but we should still investigate the performance difference.

PR #800, opened 2019-06-01
#776 .dat files don't call restore_stls in checkpoint restore

spfennell:
@alexlin0 this is a quick and dirty fix for the problem. How can I refine this so that restore_stls is called when it needs to be? @jmpenn are there any defining characteristics that separate a S_default.dat from a checkpoint or are they exactly the same? Will S_default.dat ever need to restore STLs?

### DIFF
--- a/include/trick/MemoryManager.hh
+++ b/include/trick/MemoryManager.hh
@@ -370,7 +370,7 @@ namespace Trick {
              Restore a checkpoint from the given stream.
              @param in_s - input stream.
              */
-            int read_checkpoint( std::istream* in_s, bool do_restore_stls=true);
+            int read_checkpoint( std::istream* in_s, bool do_restore_stls=false);
 
             /**
              Read a checkpoint from the file of the given name.
--- a/include/trick/MemoryManager.hh
+++ b/include/trick/MemoryManager.hh
@@ -370,7 +370,7 @@ namespace Trick {
              Restore a checkpoint from the given stream.
              @param in_s - input stream.
              */
-            int read_checkpoint( std::istream* in_s);
+            int read_checkpoint( std::istream* in_s, bool do_restore_stls=true);
 
             /**
              Read a checkpoint from the file of the given name.
--- a/trick_source/sim_services/MemoryManager/MemoryManager_restore.cpp
+++ b/trick_source/sim_services/MemoryManager/MemoryManager_restore.cpp
@@ -7,7 +7,7 @@
 #include "trick/MemoryManager.hh"
 #include "trick/ClassicCheckPointAgent.hh"
 
-int Trick::MemoryManager::read_checkpoint( std::istream *is, bool do_restore_stls /* default is true */) {
+int Trick::MemoryManager::read_checkpoint( std::istream *is, bool do_restore_stls /* default is false */) {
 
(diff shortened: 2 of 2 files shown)


trick_source/sim_services/MemoryManager/MemoryManager_restore.cpp
```diff
@@ -22,7 +22,7 @@ int Trick::MemoryManager::read_checkpoint( std::istream *is) {
     }
 
     // Search for stls and restore them
-    for ( pos=alloc_info_map.begin() ; pos!=alloc_info_map.end() ; pos++ ) {
+    for ( pos=alloc_info_map.begin() ; do_restore_stls && pos!=alloc_info_map.end() ; pos++ ) {
```
alexlin0: Move the test for do_restore_stls to the outside of the loop so it is only checked once.

include/trick/MemoryManager.hh
```diff
@@ -370,7 +370,7 @@ namespace Trick {
              Restore a checkpoint from the given stream.
              @param in_s - input stream.
              */
-            int read_checkpoint( std::istream* in_s);
+            int read_checkpoint( std::istream* in_s, bool do_restore_stls=true);
```
alexlin0: Make the default false.

trick_source/sim_services/MemoryManager/MemoryManager_restore.cpp
```diff
@@ -7,7 +7,7 @@
 #include "trick/MemoryManager.hh"
 #include "trick/ClassicCheckPointAgent.hh"
 
-int Trick::MemoryManager::read_checkpoint( std::istream *is) {
+int Trick::MemoryManager::read_checkpoint( std::istream *is, bool do_restore_stls /* default is true */) {
```
alexlin0: change comment stating default is false.

trick_source/sim_services/MemoryManager/MemoryManager_restore.cpp
```diff
@@ -60,12 +60,17 @@ int Trick::MemoryManager::read_checkpoint(const char* filename ) {
...
     // Create a stream from the named file.
     std::ifstream infile(filename , std::ios::in);
-
     if (infile.is_open()) {
-        return ( read_checkpoint( &infile ));
+
+        // If the file extention is .dat, we will tell read_checkpoint not to
+        // restore stls. It is not necessary and time consuming. 
+        bool do_restore_stls = 
+        (std::string(filename).find(".dat") != std::string::npos);
```
alexlin0: Let's make this string test specifically for "S_default.dat"

include/trick/MemoryManager.hh
alexlin0: default do_restore_stls to false

trick_source/sim_services/MemoryManager/MemoryManager_restore.cpp
alexlin0: change comment to default do_restore_stls to false

alexlin0:
This will address #526

spfennell:
Addressed your comments @alexlin0 ready for another review

Issue #805, opened 2019-06-05
Add C call to get clock_tics_per_sec.


alexlin0:
We don't have a call for this.

spfennell:
This issue will be addressed as a part of #786

Issue #786, opened 2019-05-31
Record the elapsed time of jobs executed at initialization.


alexlin0:
Record the elapsed time of all the default_data, input_processor, and initialization jobs to a file.  The file does not need to be a data recording file because each job is run only once.  Here is a not-always-correct (hardcoded sim_tic) example I recently used to print job execution times to the screen.

```c++
    while ( (curr_job = default_data_queue.get_next_job()) != NULL ) {
        long long start = clock_wall_time();
        ret = curr_job->call() ;
        long long end = clock_wall_time();
        std::cout << "Elapsed time for default data " << curr_job->name << " " << (double)(end - start)/1000000.0 << std::endl;
        if ( ret != 0 ) {
            throw Trick::ExecutiveException(ret , curr_job->name.c_str() , 0 , "default_data job did not return 0") ;
        }
    }
```

spfennell:
Will something like this work?
```cpp
/* clock_c_intf.cpp */
...
extern "C" long long clock_wall_time(void) {
    return the_clock->wall_clock_time() ;
}
...
extern "C" unsigned long long clock_tics_per_sec(void) {
	return the_clock->clock_tics_per_sec;
}
```
```cpp 
/* executive stuff */
    while ( (curr_job = default_data_queue.get_next_job()) != NULL ) {
        long long start = clock_wall_time();
        ret = curr_job->call() ;
        long long end = clock_wall_time();
        std::cout << "Elapsed time for default data " <<  curr_job->name << " "
                    << (double)(end - start)/clock_tics_per_sec() << std::endl;
        if ( ret != 0 ) {
            throw Trick::ExecutiveException(ret , curr_job->name.c_str() , 0 , "default_data job did not return 0") ;
        }
    }
```

spfennell:
#805 will be addressed as a part of this issue

PR #818, opened 2019-06-17
Add C call to clock_tics_per_sec and also log init job elapsed time

spfennell:
#786 #805


trick_source/sim_services/Executive/Executive_init.cpp
```diff
@@ -33,16 +38,30 @@ int Trick::Executive::init() {
         struct rusage cpu_usage_buf ;
         getrusage(RUSAGE_SELF, &cpu_usage_buf);
         cpu_start =   ((double) cpu_usage_buf.ru_utime.tv_sec) + ((double) cpu_usage_buf.ru_utime.tv_usec / 1000000.0);
+        std::ofstream init_log_stream;
+
+        /* First parse command line to see if trick is in Sie generation mode.
```
alexlin0: I think this would be cleaner if we called process_sim_args here and handled the sie flag in that routine.  We should remove the call to sched.process_sim_args() in share/trick/sim_objects/default_trick_sys.sm when we move it here.

If init_log_stream is processed/initialized in process_sim_args, then for cleaner code perhaps init_log_stream should be part of the class, and not passed in as an argument.

trick_source/sim_services/Executive/Executive_call_initialization.cpp
```diff
@@ -22,7 +23,12 @@ int Trick::Executive::call_initialization() {
     /* Call the initialization jobs. */
     initialization_queue.reset_curr_index() ;
     while ( (curr_job = initialization_queue.get_next_job()) != NULL ) {
+        long long start = clock_wall_time();
         ret = curr_job->call() ;
+        long long end = clock_wall_time();
+        if(init_log_stream) {
```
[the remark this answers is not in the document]
spfennell: whoops, I was trying to be fancy and use the [operator bool](http://www.cplusplus.com/reference/ios/ios/operator_bool/) for the ofstream reference, but the error bit and fail bit aren't set if no operation is called.

trick_source/sim_services/Executive/Executive_init.cpp
spfennell: @alexlin0 I have implemented this, please take another look

PR #879, opened 2019-10-03
get rid of dangling pointer use closes #878

### LINKED ISSUE #878: Changing variables with certain units does not work
A trick user noticed that trying to change the value of any variable with the unit "V" would not work and emit memory manager errors (32bit mode). This is caused by a dynamic memory bug in our Udunits interface, it may be causing other bugs with units.

spfennell:


trick_source/sim_services/VariableServer/var_server_ext.cpp
```diff
@@ -530,11 +530,12 @@ int var_set_base( const char  * var , T value , const char * units ) {
             v_tree.v_data = &v_data ;
             var_set_value( v_data , value) ;
             if ( units != NULL ) {
-                ref->units = (char *)(map_trick_units_to_udunits(units).c_str()) ;
+                ref->units = strdup((char *)(map_trick_units_to_udunits(units).c_str())) ;
             } else {
-                ref->units = (char *)units ;
+                ref->units = strdup((char *)units) ;
```
dbankieris: Both of these casts are unnecessary (and bad).

PR #892, opened 2019-10-28
Add enable flag to webserver. Default is false. Should be set to true…

jmpenn:
… in input file to enable.

### DIFF
--- a/trick_source/web/HttpServer/src/WebServer.cpp
+++ b/trick_source/web/HttpServer/src/WebServer.cpp
@@ -344,6 +344,8 @@ int WebServer::http_default_data() {
     service_websocket = false;
     shutting_down = false;
     sessionDataMarshalled = false;
+    listener = NULL;
+    enable = false;
 
     installHTTPGEThandler("vs_connections", &handle_HTTP_GET_vs_connections);
     installHTTPGEThandler("alloc_info", &handle_HTTP_GET_alloc_info);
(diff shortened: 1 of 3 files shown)


trick_source/web/HttpServer/src/WebServer.cpp
```diff
@@ -354,45 +356,51 @@ int WebServer::http_default_data() {
...
+        port = strdup(buf);
+
+        if (listener != NULL) {
+            std::cout << "Trick Webserver: Listening on port = " << port << ".\n"
+                      << "Trick Webserver: Document root = \"" << document_root << "\""
+                      << std::endl;
+        } else {
+            std::cerr << "Trick Webserver: ERROR: Failed to create listener.\n"
+                      << "Perhaps another program is already using port " << port << "."
+                      << std::endl;
```
dbankieris: Use `message_publish` instead of `cout`/`cerr`.

trick_source/web/HttpServer/src/WebServer.cpp
```diff
@@ -354,45 +356,51 @@ int WebServer::http_default_data() {
...
+            return 1;
+        }
+        mg_set_protocol_http_websocket( listener );
+        pthread_cond_init(&serviceConnections, NULL);
+        pthread_create( &server_thread, NULL, connectionAttendant, (void*)this );
     } else {
-        std::cerr << "Trick Webserver: ERROR: Failed to create listener.\n"
-                  << "Perhaps another program is already using port " << port << "."
+        std::cout << "Trick Webserver: Disabled."
```
dbankieris: Use `message_publish` instead of `cout`/`cerr`.

jmpenn: This suggestion, while a good one, doesn't have anything to do with this issue. I plan to make your suggested change, but not as part of this issue.

Issue #929, opened 2020-01-15
HELP, TRICK WONT BUILD!


spfennell:
UPDATE: [19.0.1](https://github.com/nasa/trick/releases/tag/19.0.1) and [17.5.1](https://github.com/nasa/trick/releases/tag/17.5.1) have been released to fix this issue.

As of January 15th 2020, all trick releases (17.5.0, 19.0.0, etc) can no longer download their java dependencies because of changes to security restrictions on the maven repo (see here: https://support.sonatype.com/hc/en-us/articles/360041287334). 

This has been FIXED on the master branch and on the 17 branch here at github.com/nasa/trick.

 If you need to fix your local copy you can edit ${TRICK_HOME}/libexec/trick/java/lib/Makefile and change all occurences of "http" to "https". run "make clean" in ${TRICK_HOME}/libexec/trick/java/lib to remove all of the jar files, then "make" again from the top level directory.

linked here are the changes necessary to correct this. 
https://github.com/nasa/trick/commit/909d8991f1d03cb3e163a55d5357046ea7e5444b

spfennell:
see issues #926 and #928 for more details

spfennell:
UPDATE: 19.0.1 and 17.5.1 have been released to fix this issue.

dbankieris:
If you don't care about the Java GUIs, you can run `configure` with the `--enable-java=no` option, and then `make` will complete successfully.

PR #939, opened 2020-01-31
Maven, offline mode, and --enable-mongoose

### LINKED ISSUE #938: Investigate using Maven for Java build chain
To prevent issues like #937 in the future, consider using Maven to download our JARs and build our Java GUIs.

spfennell:
pom.xml is the main project file for this update. CMake and Configure have been updated to use maven. 

Java apps are now in executable jars with all dependencies included.

closes #938


spfennell:
@ddj116

spfennell:
Worth noting that issues like #929 will be prevented by this change

PR #1017, opened 2020-07-07
Implementing PIDController to Wheelbot

avinashc99:
Files added: 
trick_sims/SIM_wheelbot/models/Control/include/PIDController.hh
trick_sims/SIM_wheelbot/models/Control/src/PIDController.cpp

Files modified:
trick_sims/SIM_wheelbot/models/Control/include/differentialDriveController.hh
trick_sims/SIM_wheelbot/models/Control/src/differentialDriveController.cpp
trick_sims/SIM_wheelbot/S_define

### DIFF
--- a/trick_sims/SIM_wheelbot/models/Control/src/PIDController.cpp
+++ b/trick_sims/SIM_wheelbot/models/Control/src/PIDController.cpp
@@ -1,0 +1,77 @@
+#include "Control/include/PIDController.hh"
+
+PIDController::PIDController(double kp, double ki, double kd, double omax, double omin, double dt, double tc) {
+  Kprop = kp;
+  Kinteg = ki;
+  Kderv = kd;
+  Dt = dt;
+  k = dt/tc;
+  error = 0.0;
+  integral = 0;
+  out_max = omax;
+  out_min = omin;
+  previous_error = 0.0;
+  prev_setpoint_value = 0.0;
+  integration_enabled = true;
+}
+
+double PIDController::getOutput(double setpoint_value, double measured_value) {
+
+  double error_unfiltered = setpoint_value - measured_value;
+
+  // Low Pass Filter
+  error = error + (k * (error_unfiltered - error));
+
+  // Check for drastic changes
+  if (prev_setpoint_value != setpoint_value) {
+    previous_error = error;
+  }
+
+  // integration
+  if (integration_enabled == true) {
+    integral = integral + (error * Dt);
+  }
+
+  // Update derivative (rate of change of error)
+  double derivative = ((error - previous_error) / Dt);
+
+  double proportionalterm = Kprop * error;
+  double integralterm = Kinteg * integral;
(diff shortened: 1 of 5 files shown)


trick_sims/SIM_wheelbot/models/Control/src/differentialDriveController.cpp
```diff
@@ -23,7 +23,14 @@ DifferentialDriveController::
     leftMotorController(LeftMotorController),
 
     rightMotorSpeedCommand(0.0),
-    leftMotorSpeedCommand(0.0)
+    leftMotorSpeedCommand(0.0),
+    desiredHeadingRate(0.0),
+
+    // PID Controller Initialization
+    headingctrl(new PIDController(1.0, 0.08, 0.5, headingRateLimit,
```
spfennell, line 30: This class needs a destructor if you are going to be allocating memory with 'new' in the constructor. Alternatively, you can change the pointers to regular objects which would remove the need for a destructor, assignment operator, copy constructor. From a glance I do not see any reason that you need to use pointers instead of objects.

trick_sims/SIM_wheelbot/models/Control/src/PIDController.cpp
```diff
@@ -0,0 +1,77 @@
...
+  double integralterm = Kinteg * integral;
+  double derivateterm = Kderv * derivative;
+  double output = proportionalterm + integralterm + derivateterm;
+
+  // Clamping Actuator
+
+  // Sign Check
+  bool same_sign;
+  if ((error * output) > 0.0) {
+    same_sign = 1;
```
spfennell, line 48: This is a C style way to write a bool, not C++. Please use 'true' or 'false' for bools or old men like me will get confused. I would let it slide, but you use 'true'/'false' elsewhere in the same file, so it needs to be consistent

trick_sims/SIM_wheelbot/models/Control/src/PIDController.cpp
```diff
@@ -0,0 +1,77 @@
...
+  } else {
+    same_sign = 0;
+  }
+
+  //Saturation Check
+  bool output_limited;
```
spfennell, line 54: output_limited MUST be initialized to 'false'. Otherwise it very well could be 'true' by default.

trick_sims/SIM_wheelbot/models/Control/src/PIDController.cpp
```diff
@@ -0,0 +1,77 @@
...
+  if (output > out_max) {
+    output = out_max;
+    output_limited = 1;
+  }
+  if (output < out_min) {
```
spfennell, line 59: Another silly little nitpick. Unless there is a feasible case in which out_max <= out min, this could be an 'if else if'.

avinashc99:
I have made changes and submitted a new pull request. I will close this one. Thank you.

Issue #1024, opened 2020-07-13
TRICK_HOST_CPU GCC Minor Version Absent On Ubuntu 18.04+


keithvetter:
A Ubuntu 18.04+ user is getting a bad TRICK_HOST_CPU with gte due to a change in gcc 7.x.  The gcc -dumpversion meaning has changed.  For gcc 7+, you want to use gcc -dumpfullversion.  gcc 4.x does not have the -dumpfullversion option.  There is trick to make it work.

The trick is to use:
gcc -dumpfullversion -dumpversion (in that order!)
instead of
gcc -dumpversion
 
inside of <trickhome>/libexec/trick/pm/gte.pm.
 
If you do that, the version will be correct for gcc 4.x and gcc 7.x+.
 
That might not be the solution you want, but it is simple.  Another way would be to do a system call to see if gcc has the -dumpfullversion option and use it if it exists, but that is more code invasive.

PR #1026, opened 2020-07-13
Fix TRICK_HOST_CPU gcc version for Ubuntu 18.04+ (#1024)

keithvetter:
A Ubuntu 18.04+ user is getting a bad TRICK_HOST_CPU with gte due to a change in gcc 7.x.  The gcc -dumpversion meaning has changed.  For gcc 7+, you want to use gcc -dumpfullversion.  gcc 4.x does not have the -dumpfullversion option.  There is trick to make it work.

The trick is to use:
gcc -dumpfullversion -dumpversion (in that order!)
instead of
gcc -dumpversion
 
inside of <trickhome>/libexec/trick/pm/gte.pm.
 
If you do that, the version will be correct for gcc 4.x and gcc 7.x+.
 
That might not be the solution you want, but it is simple.  Another way would be to do a system call to see if gcc has the -dumpfullversion option and use it if it exists, but that is more code invasive.

### DIFF
--- a/libexec/trick/pm/gte.pm
+++ b/libexec/trick/pm/gte.pm
@@ -49,21 +49,21 @@ sub gte (@) {
                 # remove possible ccache from TRICK_CC
                 $temp =~ s/.*?ccache\s+// ;
                 if ( -e $temp ) {
-                    $ret = `$temp -dumpversion` ;
+                    $ret = `$temp -dumpfullversion -dumpversion` ;
                 }
                 else {
                     printf STDERR "[33mCannot find TRICK_CC = $temp, using /usr/bin/gcc\n" ;
-                    $ret = `/usr/bin/gcc -dumpversion` ;
+                    $ret = `/usr/bin/gcc -dumpfullversion -dumpversion` ;
                 }
             } else {
                 # remove possible ccache from TRICK_CC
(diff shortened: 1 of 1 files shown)


spfennell:
We may actually need a dev branch, because it looks like pull requests from forks don't run our GitHub Actions CI jobs =/

spfennell:
@keithvetter is dumpfullversion backwards compatible? Is gcc <7.x still in use anywhere?

spfennell:
> We may actually need a dev branch, because it looks like pull requests from forks don't run our GitHub Actions CI jobs =/

You could try enabling the tests by going to the 'Actions' tab on your fork and opting in to GitHub Actions. Then you may need to make a commit to trigger the pipeline

keithvetter:
Scott, that works for both gcc 4.x and gcc 7.x.

And according to stackoverflow.com, someone was successful with using both options in the right order.

This seems to work on old and new in a uniform way:

    $ gcc -dumpfullversion -dumpversion
    4.4.7

    $ gcc -dumpfullversion -dumpversion
    7.2.1

It appears gcc ignores -dump* args it doesn't recognise, and stops processing args once it does recognise one. That's why this works, and why -dumpfullversion must be first! – Ralph Versteegen Oct 30 '19 at 13:13

PR #1052, opened 2020-09-09
1036 1007 only enable  cpp11 in Makefile_io_src when needed

### LINKED ISSUE #1007: Makefile_io_src is fixed at std=c++11
We have hardcoded Makefile_io_src to always use -std=c++11.  This is in PrintAttributes.cpp.  If a header file has c++14 or greater constructs, the io_src code will fail to compile.

### LINKED ISSUE #1036: io_src files with c++14 features may not compile
From c++ proposal [N3653](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3653.html):

```c++
// @trick_parse{everything}
class Foo {
    struct S { int a; const char* b; int c; int d = b[a]; };
    S ss = { 1, "asdf" };
};
```

Compiling on Centos 8, whose default gcc version is 8.3.1, which supports c++14 by default:
```
In file included from /home/SIM_test/S_source.hh:27,
                 from build/home/SIM_test/io_S_source.cpp:19:
/home/SIM_test/models/Foo.hh:5:21: error: could not convert ‘{1, "asdf"}’ from ‘<brace-enclosed initializer list>’ to ‘Foo::S’
  S ss = { 1, "asdf" };
```
[`PrintAttributes.cpp`](https://github.com/nasa/trick/blob/ab2cfe1a34bfe8fe8edf1098b66c5273c8fa0c03/trick_source/codegen/Interface_Code_Gen/PrintAttributes.cpp#L365) adds `--std=c++11` to `TRICK_IO_CXXFLAGS` because ICG produces code that requires at least c++11 to compile and versions of gcc on some supported distributions do not support c++11 by default. Unfortunately, that means even if the compiler supports newer versions of c++, it will only accept c++11-compliant code.

Adding `TRICK_IO_CXXFLAGS += --std=c++14` to `S_overrides.mk` solves the problem, but it would be better if ICG only specified the std when necessary.

spfennell:
closes #1036 closes #1007

### DIFF
--- a/trick_source/codegen/Interface_Code_Gen/PrintAttributes.cpp
+++ b/trick_source/codegen/Interface_Code_Gen/PrintAttributes.cpp
@@ -428,6 +428,9 @@ void PrintAttributes::printIOMakefile() {
         << "        TRICK_IO_CXXFLAGS += -std=c++11" << std::endl
         << "    endif" << std::endl
         << "endif" << std::endl
+        << "ifeq ($(IS_CC_CLANG), 1)" << std::endl
+        << "    TRICK_IO_CXXFLAGS += -std=c++14" << std::endl
+        << "endif" << std::endl
         << std::endl
         << "IO_OBJECTS =" ;
 
--- a/trick_source/codegen/Interface_Code_Gen/PrintAttributes.cpp
+++ b/trick_source/codegen/Interface_Code_Gen/PrintAttributes.cpp
@@ -420,7 +420,7 @@ void PrintAttributes::printIOMakefile() {
 
     makefile_io_src.open("build/Makefile_io_src") ;
     makefile_io_src
-        << "TRICK_IO_CXXFLAGS += -std=c++11 -Wno-invalid-offsetof -Wno-old-style-cast -Wno-write-strings -Wno-unused-variable" << std::endl
+        << "TRICK_IO_CXXFLAGS += -Wno-invalid-offsetof -Wno-old-style-cast -Wno-write-strings -Wno-unused-variable" << std::endl
         << std::endl
         << "ifeq ($(IS_CC_CLANG), 0)" << std::endl
         << "    TRICK_IO_CXXFLAGS += -Wno-unused-local-typedefs -Wno-unused-but-set-variable" << std::endl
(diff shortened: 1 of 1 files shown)


trick_source/codegen/Interface_Code_Gen/PrintAttributes.cpp
```diff
@@ -420,10 +420,16 @@ void PrintAttributes::printIOMakefile() {
...
-        << "TRICK_IO_CXXFLAGS += -std=c++11 -Wno-invalid-offsetof -Wno-old-style-cast -Wno-write-strings -Wno-unused-variable" << std::endl
+        << "TRICK_IO_CXXFLAGS += -Wno-invalid-offsetof -Wno-old-style-cast -Wno-write-strings -Wno-unused-variable" << std::endl
         << std::endl
         << "ifeq ($(IS_CC_CLANG), 0)" << std::endl
         << "    TRICK_IO_CXXFLAGS += -Wno-unused-local-typedefs -Wno-unused-but-set-variable" << std::endl
+        << "    ifeq ($(shell test $(GCC_MAJOR) -lt 6; echo $$?), 0)" << std::endl
+        << "        TRICK_IO_CXXFLAGS += -std=c++11" << std::endl
+        << "    endif" << std::endl
+        << "endif" << std::endl
+        << "ifeq ($(IS_CC_CLANG), 1)" << std::endl
```
dbankieris, line 431: Can't this just be an `else`?

trick_source/codegen/Interface_Code_Gen/PrintAttributes.cpp
```diff
@@ -420,10 +420,16 @@ void PrintAttributes::printIOMakefile() {
...
+        << "    TRICK_IO_CXXFLAGS += -std=c++14" << std::endl
```
dbankieris, line 432: Anyone can tell Trick to use clang to build all of the sim by setting `TRICK_CC` to `clang`, so we can't assume it's C++14 compatible. I sometimes do that to get better error messages when building a sim. I would do the same thing you did with gcc and add `--std=c++11`  only for versions that don't enable support for it by default.

dbankieris (commented):
Also, stop working at 2 in the morning ;-)

PR #1071, opened 2020-11-05
Trick installation on MacOS documentation update

merchantzain:
Step 5 in the installation guide “Install cask to get java and xquartz” results in the following error: “Error: Cask 'java' is unavailable: No Cask with this name exists. Did you mean one of these?”, then giving some other possible downloads. It seems that Java cannot be installed using home-brew in this way. The way around this I used was to install Java from Oracle’s website (https://www.java.com/en/download/) and the JDK from here (https://www.oracle.com/java/technologies/javase-jdk15-downloads.html), then install the other packages on their own. This has been tested to be working on macOS Catalina.

### DIFF
--- a/docs/documentation/install_guide/Install-Guide.md
+++ b/docs/documentation/install_guide/Install-Guide.md
@@ -195,13 +195,16 @@ or
 curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install | ruby
 ```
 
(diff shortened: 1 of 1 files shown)


spfennell (commented):
@merchantzain thank you for helping with our install instructions! I believe java is no longer a cask, but you can still install it with brew install java (instead of brew cask install java). Also brew install openjdk seems to work as well, I think 'java' is just a brew alias for openjdk. We may want to encourage this method as well, to better streamline the install instructions.

PR #1112, opened 2021-02-12
Add reference support to memory manager

### LINKED ISSUE #1092: Reference members are not viewable in Trick  View
References are not added to S_sie.resource and are not viewable in Trick View. Investigate and see if there is a simple solution. Because of the nature of references, this may take  a considerable effort to fix.  

```cpp
class Starter {
public:
    Starter(int bar);
    int& foo;

};
```

```cpp
Starter::Starter(int bar) foo(bar){

}

```

Screen Shot 2021-01-15 at 9 22 44 AM

dbankieris:
This PR adds support for references over the variable server and data recording, but not checkpointing.

Fixes #1092

### DIFF
--- a/trick_source/codegen/Interface_Code_Gen/FieldDescription.cpp
+++ b/trick_source/codegen/Interface_Code_Gen/FieldDescription.cpp
@@ -57,6 +57,7 @@ FieldDescription::FieldDescription(
   is_stl(0) ,
   has_stl_clear(1) ,
   is_static(0) ,
+  is_reference(0) ,
   num_dims(0) ,
   array_sizes() {} ;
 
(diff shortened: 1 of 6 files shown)


alexlin0:
OK, I can believe that references can be viewed but not set.  How are they handled for checkpoints.  Are the values written to the checkpoint file at all?

alexlin0 (approved):

dbankieris:
ICG still sets their I/O flag to exclude them from checkpointing (both reading and writing), so they continue to be absent from checkpoint files. References themselves cannot be set (the official term is "reseated"), but the object it refers to can be. The addition of the `AO_DEREFERENCE` `ADDRESS_NODE` causes the memory manager to automatically dereference references, so the value you get/set in TV is the object being referred to.

PR #1119, opened 2021-03-12
1093 c str

### LINKED ISSUE #1093: c_str() memory issues
returning a char * from the std::string::c_str() method frequently causes use-after-free/dangling pointer memory bugs when used with std::strings on the stack or temporary rvalue string copies. A trick user pointed out that this happens on  line 209 of 
 ExternalApplication_c_intf.cpp:
`return externalApplication->get_startup_command().c_str(); `

In this example, the  return value of get_startup_command() is a deep copy of the externalApplication startup command. Since the std::string is an rvalue on the stack, it gets immediately deallocated and it's c_str goes along with it, leaving a dangling pointer.  

We have encountered this issue before. Let's go ahead check  all usages of c_str() and make sure we never see this bug again.

spfennell:
closes #1093

### DIFF
--- a/include/trick/ExternalApplication.hh
+++ b/include/trick/ExternalApplication.hh
@@ -7,6 +7,7 @@
 
 #include <string>
 #include <sstream>
+#include <vector>
 
 namespace Trick {
 
--- a/include/trick/ExternalApplication.hh
+++ b/include/trick/ExternalApplication.hh
@@ -28,7 +28,7 @@ namespace Trick {
         ExternalApplication();
 
         /** Destructor. */
-        virtual ~ExternalApplication() {}
+        virtual ~ExternalApplication();
 
         /**
          * Sets #command.
@@ -43,6 +43,13 @@ namespace Trick {
          * @return the launch command
          */
         virtual std::string get_startup_command();
+
+        /**
+         * Gets #command as c str.
+         *
+         * @return the launch command as c str
+         */
+        virtual const char * get_startup_command_c_str();
 
         /**
          * Appends #arguments.
@@ -216,9 +223,6 @@ namespace Trick {
         /** Designates from which source an option is to be used. */
         enum OptionSource {AUTO, USER, NONE};
 
-        /** Command to execute when starting this application. */
-        std::string command;
-
         /** Arguments to be passed after the command. */
         std::ostringstream arguments;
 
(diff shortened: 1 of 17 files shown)


abrogley:
I think this introduced new warnings in the SIM_Ball++_L1 sim, perhaps in others as well.

```
abrogley@machine:~/trick/trick-fork/trick_sims/SIM_Ball++_L1$ ./S_main_Linux_7.5_x86_64.exe RUN_test/input.py
MemoryManager:WARNING:(None:0): Couldn't find an io_src_sizeof_ function for type void[io_src_sizeof_void()].

MemoryManager:WARNING:ATTRIBUTES for type "void" not found.

MemoryManager:ERROR:get_type_attributes failed for type: void


MemoryManager:WARNING:(None:0): Couldn't find an io_src_sizeof_ function for type void[io_src_sizeof_void()].

MemoryManager:WARNING:ATTRIBUTES for type "void" not found.

MemoryManager:ERROR:get_type_attributes failed for type: void


MemoryManager:WARNING:(None:0): Couldn't find an io_src_sizeof_ function for type void[io_src_sizeof_void()].

MemoryManager:WARNING:ATTRIBUTES for type "void" not found.

MemoryManager:ERROR:get_type_attributes failed for type: void


MemoryManager:WARNING:(None:0): Couldn't find an io_src_sizeof_ function for type void[io_src_sizeof_void()].

MemoryManager:WARNING:ATTRIBUTES for type "void" not found.

MemoryManager:ERROR:get_type_attributes failed for type: void


MemoryManager:WARNING:(None:0): Couldn't find an io_src_sizeof_ function for type void[io_src_sizeof_void()].

MemoryManager:WARNING:ATTRIBUTES for type "void" not found.

MemoryManager:ERROR:get_type_attributes failed for type: void


|L   1|2021/03/12,13:52:04|machine| |T 0|0.000000| Dumped ASCII Checkpoint chkpnt_pre_init.
|L   0|2021/03/12,13:52:04|machine| |T 0|0.000000| time =     0.00 , position =     5.000000 ,     5.000000
|L   0|2021/03/12,13:52:04|machine| |T 0|10.000000| time =    10.00 , position =    -0.168112 ,     3.484156
|L   0|2021/03/12,13:52:04|machine| |T 0|20.000000| time =    20.00 , position =    -3.103971 ,    -4.629418
```

PR #1142, opened 2021-05-03
Trickops

spfennell:
Get trickops history


ddj116:
@spfennell just an FYI, I don't think we have any reason to maintain this PR anymore. I think you opened it to preserve the commit history, but we successfully stepped up to latest trick to get TrickOps and no longer need those refs. Additionally, now that I can push topic branches directly to the this project, the CI complexities of referencing my personal fork goes away entirely 👍

PR #1145, opened 2021-05-17
Zip SWIG-generated python modules

### LINKED ISSUE #1144: Native support for zipping the swig generated python files
## Motivation

During simulation startup, the python input processor `import`s a lot of files that are not in user space but are generated by swig under the `SIM_*/trick/` directory. For small sims, these are just a handful of files, but for large sims, like in the `ramtares/` workflow for example, these are thousands of files. Each time a single sim instance starts, the sim must interact with the file system to open all of these files before it even gets to the first executable line of `RUN_*/input.py`. 

On local filesystems, this process happens quite quickly, but Monte-carlo workflows typically necessitate runs be executed in a workspace on a network file system, since runs are not all run on the same machine and copying sims around is cumbersome. This usually means dozens, hundreds, and sometimes thousands of runs all start up near the same wall-clock time, and when each sim must read thousands of files from the network file system, this means potentially a million file open operations hitting the same network all at once.

We have seen extreme network filesystem slowness in AGDL and now FSL consistently over the years, and a recent evaluation of the NFS metrics combine with `strace` on the sim leads us to believe it is these massive amount of file `open()`s on the network FS that are contributing to the network slowness.

## Reducing the file I/O by zipping `SIM_*/trick/`

It has recently come to my attention that python understands how to read modules from a `.zip` file natively, and I recently prototyped this approach to replace all of our 3000 file reads with a single read of `trick.zip`. Basically this is how it works:

```sh
# cd path/to/SIM, build the sim, then...
python -m compileall trick/             # Build all the .pyc files
zip  trick.zip trick/*.{py,pyc}         # zip 'em up
rm -rf trick/                           # Remove the original directory
# Now run, using the zip file
PYTHONPATH="trick.zip" ./S_main... RUN.../input.py
```
Confirmed via `export PYTHONPATH="trick.zip"; strace -f -e trace=%file ./S_main... RUN.../input.py`, all individual file reads from `trick/` are replaced with the single read of `trick.zip`.

## Complications associated with Trickified libraries -- needing to use `PYTHONPATH`

The zip by itself isn't enough, you have to adjust `sys.path` so that the `python` interpreter can import the modules under the zipped directory. Although `TRICK_PYTHON_PATH` can typically be adjusted in `S_overrides.mk` to add the zipped area, that approach does not get the entry into `sys.path` early enough for Trick to find it in the case where a [Trickified library's](https://nasa.github.io/trick/documentation/building_a_simulation/Trickified-Project-Libraries) python files import trick modules like `import sim_services`.  In this scenario, the sim will die on the first input processor line that tries to reference a sim object.

@dbankieris was able to help figure out why this happens, here's his explanation copied from our discussion:

> Here's what's going on. `IPPython.cpp` adds the current working directory (`SIM_*`) to `sys.path`, then `trick/pymods`, then everything in `TRICK_PYTHON_PATH`. It then imports `trick`, which is found under `SIM_*`. `SIM_*/trick/__init__.py` adds `SIM_*/trick` to `sys.path`, which allows imports like `import sim_services`. When we replace trick with `trick.zip`, the `__init__.py` within still adds `SIM_*/trick` to `sys.path`, whereas it now needs to add `SIM_*/trick.zip/trick` instead. If you set `PYTHONPATH = SIM_*/trick.zip:SIM_*/trick.zip/trick`, then everything works without needing to modify any import statements or Trick itself. As a separate issue, the reason you have to use `PYTHONPATH` and not `TRICK_PYTHON_PATH` is because `trick/pymods` is added before the contents of `TRICK_PYTHON_PATH`, and there is a trick module in both `trick/pymods` and `trick.jar`, which prevents the zipped `__init__.py` from running, since the trick module is found in `pymods` first, whereas it would usually be found in `SIM_*` first. I think the long-term solution is for Trick to produce `trick.zip` and add `trick.zip/trick` to `sys.path`.

## The long term solution

In ramtares we are rolling our own solution by modifying our environment `PYTHONPATH` and `S_overrides.mk` to manage the creation of `trick.zip`. But it would be nice if this mechanism was native to Trick. Here I'd like to discuss:

* Is moving from `SIM*/trick/` --> `SIM*/trick.zip` a change that should apply globally? Should it be an option to Trick? Are there any downsides to ditching the original approach?
* Can/should this be applied to the framework that Trickifies libraries as well? 

FYI @jmpenn @spfennell @alexlin0 @dbankieris

dbankieris:
Closes #1144


share/trick/makefiles/trickify.mk
```diff
@@ -46,8 +46,8 @@
 # trickified.o. You should choose something more meaningful, especially if
 # you're using another build type.
 #
-# TRICKIFY_PTYON_DIR (optional)
-# The directory into which generated Python modules are placed. The default
+# TRICKIFY_PTYON_ZIP (optional)
```
ddj116, line 49: Looks like a typo on `PYTHON`, doesn't match the documentation above.

share/trick/makefiles/trickify.mk
```diff
@@ -86,15 +86,29 @@
...
 TRICKIFY_OBJECT_NAME ?= trickified.o
+
+# We started zipping the Python modules, and the original variable is now misnamed.
+# For backward compatibility, initialize the new variable from the old variable if set, and then
+# use the old vairable for something else! python.zip would be a better default value, but leaving
+# it as python will allow existing Trickified projects that don't override the default to work both
+# before and after this change.
 TRICKIFY_PYTHON_DIR ?= python
+# Use origin instead of ?=, which defaults to recursive expansion, since we change
+# TRICKIFY_PYTHON_DIR after this assignment and don't want that affecting TRICKIFY_PTYHON_ZIP.
```
ddj116, line 105: I've read `PYTHON` and `PTYHON` so many times now, I don't know which is correct!

libexec/trick/make_makefile_swig
```diff
@@ -218,6 +218,14 @@ SWIG_OBJECTS = \$(subst .cpp,.o,\$(SWIG_SRC)) $swig_src_dir/init_swig_modules.o
 \$(S_MAIN): \$(SWIG_OBJECTS)
 
 LINK_LISTS += \$(LD_FILELIST)build/py_link_list
+
+# trick.zip ====================================================================
+
+trick.zip: \$(SWIG_SRC)
+\t\$(call ECHO_AND_LOG,zip -rq trick trick)
+\t\$(call ECHO_AND_LOG,rm -rf trick)
```
ddj116, line 226: I don't see the rule to pre-compile all the `*.pyc` files, is it somewhere else?

dbankieris:
@ddj116 I think this is ready now. Can you give it a spin before I merge?

ddj116:
Confirmed this MR works fine with my gateway project -- we'll move to it as soon as it's in master. Antares has to wait for other dependency reasons unrelated to these changes, so we'll continue to "roll our own zip" in that project until we are able to step up to the latest Trick.

Approved and much appreciated!

dbankieris:
@alexlin0 Dan and I thought it might be helpful to rename `SIM_*/trick` to `SIM_*/.trick` (a hidden directory) as an indication to users that things have changed and modifying the non-zipped Python modules will have no effect on the sim. Any objection?

alexlin0:
No objection to the move.

Issue #1155, opened 2021-06-02
Fix "Argument list too long" error for large trickified libraries and/or long build paths


ddj116:
## The problem

When a trickified library has too many `$(IO_OBJECTS)`, `$(SWIG_OBJECTS)`,  or the build path is of sufficiently large length, this error can occur on the final link  of `$(TRICKIFY_OBJECT_NAME)`:
```
Linking^[[0m    trickified_lib.o
make[1]: execvp: /bin/bash: Argument list too long
```

This has been replicated for the JEOD trickified library in ramtares with an extremely long path to the repository

## The fix

We can change `ld` to use `@file` notation for input which will shorten the recipe length as run in the shell

dbankieris:
Note that `ld` is only used when `TRICKIFY_BUILD_TYPE` is `PLO`. Solutions for `TRICK_CXX` and `ar` are also needed.

dbankieris:
Oh, they all support `@file`. Woohoo!

PR #1156, opened 2021-06-02
Change ld to use @file input to shorten recipe length in shell for trickified libs

ddj116:
Fixes the "execvp: /bin/bash: Argument list too long" error for
sufficiently large trickified libraries / absolute build paths

Refs #1155

### DIFF
--- a/share/trick/makefiles/trickify.mk
+++ b/share/trick/makefiles/trickify.mk
@@ -119,14 +119,20 @@ TRICK_EXT_LIB_DIRS :=
 .PHONY: all
 all: $(TRICKIFY_OBJECT_NAME) $(TRICKIFY_PYTHON_DIR)
 
-$(TRICKIFY_OBJECT_NAME): $(SWIG_OBJECTS) $(IO_OBJECTS) | $(dir $(TRICKIFY_OBJECT_NAME))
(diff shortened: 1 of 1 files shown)


ddj116:
Ack although the library builds the library is missing content:

```
init_swig_modules.cpp:7396: undefined reference to `init_m45f01a5b169e22ab307e751f71bae1f4'
```
Investigating...

share/trick/makefiles/trickify.mk
```diff
@@ -119,14 +119,20 @@ TRICK_EXT_LIB_DIRS :=
 .PHONY: all
 all: $(TRICKIFY_OBJECT_NAME) $(TRICKIFY_PYTHON_DIR)
 
-$(TRICKIFY_OBJECT_NAME): $(SWIG_OBJECTS) $(IO_OBJECTS) | $(dir $(TRICKIFY_OBJECT_NAME))
+$(BUILD_DIR)S_library_swig_objects: $(BUILD_DIR)S_library_swig
```
ddj116, line 122: Although this works in solving my specific use-case, these `echo ... | tr ...` lines could themselves result in `Argument list too long` if the library / build path was long enough. Without being able to use `$(file ...)` I don't know of a better way to do this, if you have suggestions let me know.

ddj116:
Closing in favor of more robust implementation in #1157

PR #1168, opened 2021-08-03
Rewrite the webserver to use the CivetWeb library instead of Mongoose.

cherpin00:


jmpenn:
I don't believe the changes to  trick_sims/Ball/SIM_ball_L1/RUN_test/input.py are relevant here.

jmpenn:
The following two files shouldn't be part of this pull request:
- trick_sims/Cannon/SIM_cannon_webserver/civet_server_error.log
- trick_sims/Cannon/SIM_cannon_webserver/log.error

Also I think should delete:
- trick_sims/Cannon/SIM_cannon_webserver/server.pem
but make sure a user knows that they need to generate their own server.pem

jmpenn:
This is for visual studio?
trick_source/web/CivetServer/.vscode/settings.json

cherpin00:
> This is for visual studio?
> trick_source/web/CivetServer/.vscode/settings.json

Yes, I will remove this file.  I don't think it belongs in the pull request.

Issue #28, opened 2015-03-25
Isolate DMTCP code


alexlin0:
There are ifdefs for DMTCP scattered throughout the code.  Consolidate all of these into the DMTCP directory.

alexlin0:
Action: Ask if users are using DMTCP.  If not changed this issue to delete all DMTCP support code.

PR #1183, opened 2021-08-28
Initial excision of DMTCP from Trick. #28

jmpenn:
This completely removes DMTCP from Trick.


autoconf/configure.ac
```diff
@@ -571,13 +527,12 @@ AC_SUBST([UDUNITS_EXCLUDE])
...
 AX_GTEST_HOME([])
 AX_ER7_UTILS([])
 
-AC_OUTPUT([share/trick/makefiles/config_user.mk])
+AC_CONFIG_FILES([share/trick/makefiles/config_user.mk])
+AC_OUTPUT
+
+
 
-printf "\033@<:@32mconfigure script successfully completed\033@<:@0m\n"
```
spfennell, line 583: Is there a reason for removing this print statement?

trick_source/java/src/main/java/trick/simcontrol/SimControlApplication.java
```diff
@@ -995,13 +971,8 @@ private JPanel createCommandsPanel() {
...
         GridLayout gridLayout = null;
-        if (!isDmtcpOptionOn) {
-        	// 2 columns and 5 rows, each component has the same width and height.
-            gridLayout = new GridLayout(5,2,2,4);
-        } else {
-        	// 2 columns and 6 rows
-            gridLayout = new GridLayout(6,2,2,4);
-        }
+        // 2 columns and 5 rows, each component has the same width and height.
+        gridLayout = new GridLayout(5,2,2,4);
```
spfennell, line 975: Suggestion: 
replace
```java
GridLayout gridLayout = null;
gridLayout = new GridLayout(5,2,2,4);
```
with
```java
GridLayout gridLayout = new GridLayout(5,2,2,4);
```

spfennell (commented):
It might be a good idea to test Sim Control Panel manually and make sure there isn't any layout problems before merging.

autoconf/configure.ac
jmpenn: Nope! There is not good reason for removing the print statement. Forgot to add it back in. I'll do that.

Issue #1189, opened 2021-09-14
Fix ICG errors while processing system header files


alexlin0:
ICG prints a lot of errors it finds in system header files.  These are caused by ICG not having the clang preprocessor configured correctly.

PR #1190, opened 2021-09-14
Fix ICG errors while processing system header files #1189

alexlin0:
Found an InitPreprocessor call and am using it.  Not sure which version
of clang it was added, just using for the version I'm on (10) and above.
Also found some GCC defines that are used during normal compilation.
Added these to our list of defines during ICG.  This clears up all of
the errors I've been seeing.


dbankieris:
A little Googling suggests these defines reflect architecture-specific capabilities. I'm not sure hard-coding them is a good idea.

alexlin0:
I need to figure out where they get defined and extract them properly.

alexlin0:
OK, found the flag I needed.  This flag was in clang 10, but not in 9.  I did not check to see if it is in 11 or 12.

PR #1193, opened 2021-10-07
Always requeue timed-out MonteCarlo runs

### LINKED ISSUE #1192: MonteCarlo jobs that timeout can be lost forever
When the master determines that a slave has timed out, [`MonteCarlo::check_timeouts`](https://github.com/nasa/trick/blob/0b373d96cf0fdba1b3c1696d7d312a4fb29f7b41/trick_source/sim_services/MonteCarlo/MonteCarlo_funcs.cpp#L383) sets the slave's state to `MonteSlave::MC_UNRESPONSIVE_RUNNING`  and requeues the run for dispatching. If that slave later returns results, [`MonteCarlo::handle_run_data`](https://github.com/nasa/trick/blob/0b373d96cf0fdba1b3c1696d7d312a4fb29f7b41/trick_source/sim_services/MonteCarlo/MonteCarlo_receive_results.cpp#L84) removes the run from the queue if present. If the slave's result is that the run timed out, we ought to put the run back in the queue. However, the `MonteRun::MC_RUN_TIMED_OUT` case does not requeue the run if the slave's state is `MonteSlave::MC_UNRESPONSIVE_RUNNING`, which it will be if the master has already determined that this slave has timed out.

Therefore, if a timed-out run (as determined by the master) is not re-dispatched before the original slave reports a timeout, then that run is removed from the queue and lost forever.

The state check in the `MonteRun::MC_RUN_TIMED_OUT` case probably should have been removed with 27bf0b030bf97c886de1350a5c58dc6e03085711. This will result in the run being unnecessarily requeued if the master previously detected this timeout and has already re-dispatched the run to another slave, but in such a case the redundant results will be discarded by [`handle_run_data`](https://github.com/nasa/trick/blob/0b373d96cf0fdba1b3c1696d7d312a4fb29f7b41/trick_source/sim_services/MonteCarlo/MonteCarlo_receive_results.cpp#L91).

dbankieris:
Closes #1192

@tfleming-osr How's this look?

### DIFF
--- a/trick_source/sim_services/MonteCarlo/MonteCarlo_receive_results.cpp
+++ b/trick_source/sim_services/MonteCarlo/MonteCarlo_receive_results.cpp
@@ -81,7 +81,7 @@ void Trick::MonteCarlo::handle_run_data(Trick::MonteSlave& slave) {
     }
 
     /**
-     * <ul><li> Try to remove this run from the queue in case it was requeue by #check_timeouts.
+     * <ul><li> Try to remove this run from the queue in case it was requeued by #check_timeouts.
      * This covers the case in which the master determines that a slave has timed out, requeues
      * the run, and then the slave reports results.
      */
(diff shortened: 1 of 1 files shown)


tfleming-osr:
Yep, this is exactly how I fixed it in my local experiments.

It feels a little odd to dequeue the run and then requeue it again, but changing that would require reordering operations in the function so that the return value is known prior to dequeuing. So I'm all for merging this as-is.

PR #1222, opened 2022-02-02
1219 reference ignored in name from address

### LINKED ISSUE #1219: ref_name_from_address doesn't account for reference variables correctly
See the following example sim that shows how the call to ref_name_from_address() does not return the correct name if the class contains a reference.

[ref_name_from_address_fail.zip](https://github.com/nasa/trick/files/7959692/ref_name_from_address_fail.zip)

From the debugger it looks like the code that recursively looks through the attributes for a matching name for an address is using the references typed-size `sizeof(Foo)` instead of the size of a reference `sizeof(Foo &)` when checking if the address is contained by the named parameter.

spfennell:
closes #1219


spfennell:
closes #1223 closes #1224

Issue #1250, opened 2022-03-22
Disambiguate python use


alexlin0:
Several scripts use /usr/bin/python or other variations that are not explicit to which python version to use. Time to move into the present and set all of the scripts to python3.

spfennell:
Anything in Trick calling python should be using the PYTHON variable set by the configure script, which will generally be python3 on any recent OS releases.

spfennell:
> Anything in Trick calling python should be using the PYTHON variable set by the configure script, which will generally be python3 on any recent OS releases.

To clarify, I think this is the fix for this issue

spfennell:
@alexlin0 Oh I see, you were changing the #!. Nevermind!

PR #1251, opened 2022-03-22
Disambiguate python use #1250

alexlin0:
Changed all calls to /usr/bin/python to /usr/bin/env python3.  Removed execute permissions on a lot of files that are not executable.


spfennell:
Looks like you may have deleted some SIM_lander and SIM_balloon code?

Issue #1256, opened 2022-04-25
Trick trying to process anonymous enums on LLVM 13 when it shouldn't


spfennell:
OS: macOS Monterey
LLVM: 13.0.1
```
class foo {
public:
  enum {BAR = 1};
}
```

enum gets a weird name '9)' which cannot and should not compile. 

anonymous enum should not be processed.

PR #1257, opened 2022-04-25
iss #1256 add new enum type description identifier 'unnamed' to strin…

spfennell:
…gs to look for when identifying anonymous enum


spfennell:
Here is the relevant llvm commit:
https://github.com/llvm/llvm-project/commit/50542d504dd869fe1241a9cc987b72ead5a56073

Issue #1323, opened 2022-07-27
Test coverage for the variable server


jdeans289:
Currently, the variable server does not have any unit tests, and the one integration test that uses it (SIM_test_varserv) is broken.

Issue #1311, opened 2022-07-21
Add new command to variable server to send a variable only once


jdeans289:
Currently, the default behavior of the variable server is to send a list of variables to the client at a set frequency. To get a variable only once, as is common during setup, programmers have to do a var_add, var_send, var_clear combination for every desired parameter. Add a new command to the variable server that only sends a response once. The response should use a different message type than the normal repeated variable messages in order to allow simultaneous use.

PR #1330, opened 2022-08-09
Variable Server - var_send_once and integration test

jdeans289:
Makes several changes relating to the variable server:
- Adds the send_once command and message type, which allows a user to request a variable to be sent immediately and only once (intended to replace the var_add, var_send, var_clear idiom that is commonly used for this purpose)
- Minor refactoring of variable server internals to reduce repeated code
- Adds SIM_test_varserv to integration test to test basic variable server functionality
- Changes graphics client for SIM_billiards to use var_send_once as an example of intended use


trick_source/sim_services/VariableServer/VariableServerThread_copy_sim_data.cpp
```diff
@@ -92,14 +93,11 @@ int Trick::VariableServerThread::copy_sim_data() {
...
+        if (cyclical) {
+            var_data_staged = true;
+            packets_copied++ ;
+        }
 
         pthread_mutex_unlock(&copy_mutex) ;
     }
-
-
-    return (0) ;
```
spfennell, line 102: We need this return 0. Double checked the build and we now get a warning from this.

include/trick/VariableServerThread.hh
```diff
@@ -426,9 +449,19 @@ namespace Trick {
             int transmit_file(std::string file_name);
 
             /**
-             @brief Called by write_data to write data to socket in var_binary format.
+             @brief Called by write_data to write given variables to socket in var_binary format.
+            */
+            int write_binary_data( int Start, char *buf1, const std::vector<VariableReference *>& givenVars, VS_MESSAGE_TYPE messageType);
```
spfennell, line 454: I know Trick is inconsistent with variable names, but let's go ahead and call ```givenVars``` ```given_vars``` unless there's a good reason to use camel case here.

test/SIM_test_varserv/models/varserv/src/VS_tests.cpp
```diff
@@ -33,112 +33,168 @@ int VSTest::strcmp_IgnoringWhiteSpace(const char* s1, const char* s2) {
     }
 }
 
-
+// Not really sure what this is supposed to be testing, should probably just remove
```
spfennell, line 36: Yes, you can remove anything in the VS tests that you don't think is meaningful.

trick_source/sim_services/VariableServer/VariableServerThread_commands.cpp
```diff
@@ -46,17 +46,18 @@ REF2* Trick::VariableServerThread::make_error_ref(std::string in_name) {
     return new_ref;
 }
 
-int Trick::VariableServerThread::var_add(std::string in_name) {
-
-    VariableReference * new_var ;
+Trick::VariableReference* Trick::VariableServerThread::create_var_reference(std::string in_name) {
```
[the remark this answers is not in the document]
spfennell: Although I'm reading about compilers and now I think we can probably just let the compiler inline it if it wants to.

trick_source/sim_services/VariableServer/VariableServerThread_commands.cpp
```diff
@@ -93,6 +99,36 @@ int Trick::VariableServerThread::var_add(std::string var_name, std::string units
     return(0) ;
 }
 
+// TODO: This should go somewhere else probably?
```
spfennell, line 102: It's used only here, personally I'd leave it. Maybe add a comment saying ```//Helper function for var_send_once``` so we can see that it's just for this file at a glance

trick_source/sim_services/VariableServer/VariableServerThread_commands.cpp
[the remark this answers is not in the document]
jdeans289: Left as is

trick_source/sim_services/VariableServer/VariableServerThread_copy_sim_data.cpp
jdeans289: Fixed. I'm not sure why I took that out, probably an accident.

include/trick/VariableServerThread.hh
jdeans289: Fixed this and some other camel case variables that I saw

trick_source/sim_services/VariableServer/VariableServerThread_commands.cpp
jdeans289: Added

test/SIM_test_varserv/models/varserv/src/VS_tests.cpp
jdeans289: Removed this and some other commented out code in VS test

jdeans289:
closes #1323 
closes #1332 
closes #1311

Issue #1335, opened 2022-08-23
RHEL8 std::string arrays do not restore properly.


excaliburtb:
See attached for example SIM_test, a checkpoint and a restart run using Trick 19.4.0

On CentOS 7, this checkpoint/restarts just fine. On RHEL8-based OS's, it segfaults during restart.

[SIM_string_restart.zip](https://github.com/nasa/trick/files/9410936/SIM_string_restart.zip)

Simply build, run the RUN_chkpt run, then RUN_restart

PR #1336, opened 2022-08-24
Fix issue with writing to an array of std::strings.

excaliburtb:
Fixes:
https://github.com/nasa/trick/issues/1335


jmpenn:
Thanks Thomas.

PR #1341, opened 2022-09-12
Update SIM_webpanel with new functionalities 

zeronelson:
I have added the functionality to Freeze At/In, create checkpoint objects, and change real time ratio and font type.


jmpenn:
I have several questions about this pull request :
1) It contains a modification to SIM_lander. Why?
2) it adds log.err in SIM_cannon_webserver. Don't think we want to add log.err.
3) Why duplicate SIM_balloon?  Why not just add the panel updates to SIM_balloon?

jmpenn:
There are 1348 files in this merge request. Apparently you've added all of the civet web source. We don't want to do that.

jmpenn (commented):
Of the 39 files in this pull request, I only see 11 that should be here:
* trick_sims/SIM_balloon/RUN_test/input.py
* trick_sims/SIM_balloon/S_define
* trick_sims/SIM_balloon/models/graphics/src/main/java/trick/ballonDisplay/BalloonDisplay.java
* trick_sims/SIM_balloon/www/example/helloworld.html
* trick_sims/SIM_balloon/www/example/helloworld.js
* trick_sims/SIM_balloon/www/example/helloworld.json
* trick_sims/SIM_balloon/www/images/trick.png
* trick_sims/SIM_balloon/www/images/trick_icon.png
* trick_sims/SIM_balloon/www/src/style.css
* trick_sims/SIM_balloon/www/src/wsexp.css
* trick_sims/SIM_balloon/www/src/wsexp.html

The remaining 28 files are not appropriate for this pull request.

The pull request should never contain .class files.

Also, the Aircraft display should be submitted in a completely different pull request.

PR #1364, opened 2022-10-10
Add a way to disable sim objects

### LINKED ISSUE #1217: no executive function to see if a sim object is disabled. Disabling a sim object doesn't actually disable it, it just disables all the jobs.
@alexlin0 @jmpenn 

Problem: user pointed out there is no easy way to check if a sim object has been "turned off" or "disabled". This exposed a larger problem. 

set_sim_object_onoff disables/enables all jobs in a sim object. It does not set a bool in the SimObject or anywhere else as far as I know. These jobs can also be individually enabled, even if the sim object has been "turned off". 

This is some nice flexibility but could cause headaches for users who might accidentally enable a job for a sim object that has been disabled. Also (more importantly) if a user disables a job, then disables a sim object, and then reenables that sim object, the job will no longer be disabled. This feels wrong to me. 

Proposed design:
If a sim object is disabled none of its jobs will run, but they will retain their disabled/enabled status in the case that the sim object is reenabled. This has the added benefit that you can make a simple call to see if the sim object is currently disabled. 

We will add a function(s) that also disables/enables all jobs of a sim object, but doesn't disable the sim object itself (this retains the behavior of the old function).

jdeans289:
Adds functions that differentiate disabling jobs in a sim object from disabling the sim object itself, along with tests and documentation.

```
// Command to turn on/off the sim_object. If sim object is disabled then enabled, jobs retain their status from before.
int set_sim_object_onoff(std::string sim_object_name, int on) ;

// Command to get sim object on/off status. 1 for on, 0 for off, -1 if object not found.
int get_sim_object_onoff(std::string sim_object_name);

// Command to turn all the individual jobs on/off. Does not set overall sim object status.
int set_sim_object_jobs_onoff(std::string sim_object_name, int on) ;

```

closes #1217


include/trick/SimObject.hh
```diff
@@ -97,12 +106,12 @@ namespace Trick {
             /**
              * Enables all jobs in the SimObject
              */
-            void enable() ;
+            void enable_all_jobs() ;
```
jdeans289, line 109: Would this change break some existing sims?

trick_source/sim_services/Executive/test/Makefile
```diff
@@ -27,8 +30,26 @@ all : $(TESTS)
...
 
+code-coverage: test
+	# Give rid of any old code-coverage HTML we may have.
+	rm -rf lcov_html
+	# Gather coverage information about the src code.
+	lcov --capture \
+	--directory ../object_${TRICK_HOST_CPU} \
+	--base-directory ../ \
+	--output-file src_coverage.info
+	# Filter out information about directories that we don't care about.
```
ArjunSharda, line 41:
```suggestion
	# Remove information about directories that we don't care about.
```

ArjunSharda (changes requested):
Hi! Nice PR. I have a suggestion though, that I made.

jmpenn (approved):
Looks like a solid change to me. I like the overall completeness: the testing, the depreciation messages.

PR #1378, opened 2022-11-01
Test and Document interaction between checkpointing and data recording

### LINKED ISSUE #1371: Investigate, test, and document interactions between checkpointing and data recording
We have been getting lots of questions recently about how checkpointing and data recording do and should interact. This issue should answer the following questions:
1. Are data recording groups dumped in a checkpoint?
2. Does a checkpoint restore also restore and re-initialize data recording groups?
3. If a checkpoint that includes data recording is loaded into a sim that was already recording something, what happens to the data recording file?
4. Can trick reliably load checkpoints containing data recording that were dumped before and after the current sim time? (e.g. load a checkpoint at t=5 that was dumped when t=2, and vice versa)

1. The answer to all of the above questions is clearly documented in user documentation.
2. A detailed design writeup of checkpoint and data recording interaction is created for developer documentation.
3. Integration test(s) that display the correct behavior in all cases have been created and are passing.
5. Any bugs found during this process are resolved.

jdeans289:
* Add test sim for checkpointing with data recording, SIM_checkpoint_data_recording
* Document how checkpoints and data recording are expected to interact


jdeans289:
It looks like getting TrickOps to play nice with the CI pipeline will take some doing. Created a new issue to track this problem in #1380

jdeans289:
The decrease in coverage is probably because I accidentally committed SIM_amoeba commented out

jdeans289:
It seems like one of the tests has revealed a race condition in checkpointing, which causes a segmentation fault when the checkpoint is reloaded. It appears that sometimes the variable name and/or variable alias in the data recording group does not get checkpointed correctly.

```
char* trick_anon_local_8[1];
char* trick_anon_local_9[1];
char trick_anon_local_10[23];
char trick_anon_local_11[1];
...
```

In a correct checkpoint:
```
trick_anon_local_8 = 
    {&trick_anon_local_10[0]};

trick_anon_local_9 = 
    {&trick_anon_local_11[0]};

```

In a buggy checkpoint:
```
trick_anon_local_8 = 
    {&trick_anon_local_10[0]};

trick_anon_local_9 = "";

```
For some reason, the variable_alias (sometimes variable_name as well, but less common) is not getting hooked up to the correct reference. This happens in about 1/30 runs, and only on RUN_test6

jdeans289:
Closing since a lot of TrickOps work here was done separately. The actual data record checkpointing test has been cherry picked out into a clean branch.

PR #1383, opened 2022-11-04
Stop attaching a debugger on segfault on Mac

### LINKED ISSUE #1382: Remove code that attaches a debugger on segfault in Mac
Trick has code in sim_services/Executive/Executive_signal_handler.cpp:73 that attaches a debugger and tries to produce a stack trace when a segfault signal is received. On Mac, this does not work. Maybe this worked at some point, but with the recent Mac security changes, this now will just hang, and will sometimes even crash the terminal and require a restart. Since the only effect it has now is slowing down development, it should be removed.

1. This code is machete'd

jdeans289:
closes #1382


jdeans289:
Looks like something's up with the rockylinux package manage O_o

PR #1390, opened 2022-11-08
Refactored sim tests to run through TrickOps

jdeans289:
- Modifies Github Actions workflows to be nice to TrickOps
- Run all sim tests through TrickOps
- Adds an optional 'binary' argument to configuration yml for TrickOps (so that I can use T_main rather than S_main)


jdeans289:
@ddj116

share/trick/trickops/TrickWorkflow.py
```diff
@@ -390,11 +391,19 @@ def type_expected(var, expected_type, fatal=False, extra_msg=''):
             sanitized_labels =( [l for l in c[s]['labels'] if type_expected(
               l, expected_type=str, extra_msg='Ignoring label "%s".' % l) ] )
             self.config[s]['labels'] = sanitized_labels
+          # Check for binary argument
+          if 'binary' not in c[s] or not c[s]['binary']:
+            self.config[s]['binary'] = self.defaults['binary']
+
+          # Add the CPU format string (won't fail if there is no CPU placeholder)
+          self.config[s]['binary'] = self.config[s]['binary'].format(cpu=self.trick_host_cpu)
```
ddj116, line 399: Nice!

trickops.py
```diff
@@ -0,0 +1,42 @@
...
+        # Create the trick_test directory if it doesn't already exist
+        if not os.path.exists(trick_top_level + "/trick_test"):
+          os.makedirs(trick_top_level + "/trick_test")
+
+        # Base Class initialize, this creates internal management structures
+        TrickWorkflow.__init__(self, project_top_level=(trick_top_level), log_dir=(trick_top_level +'/trickops_logs/'),
+            trick_dir=trick_top_level, config_file=(trick_top_level + "/test_sims.yml"), cpus=3, quiet=True)
+    def run( self ):
+
+      # Change all the build commands to have the whole path
```
ddj116, line 18: I ran into something similar yesterday when I realized that `${TRICK_HOME}/bin/` has to be on your `PATH` or sim builds will fail with `trick-CP: command not found`. Never ran into this in my Gateway usage of TrickOps because our project environment takes care of it.  Here's what I did as a quick hack, what do you think?

```diff
diff --git a/share/trick/trickops/TrickWorkflow.py b/share/trick/trickops/TrickWorkflow.py
index 8ab5076..4ebb535 100644
--- a/share/trick/trickops/TrickWorkflow.py
+++ b/share/trick/trickops/TrickWorkflow.py
@@ -100,9 +100,6 @@ class TrickWorkflow(WorkflowCommon):
             available
         """
         super().__init__(project_top_level=project_top_level, log_dir=log_dir, quiet=quiet)
-        # If not found in the config file, these defaults are used
-        self.defaults = {'cpus': 3, 'name': None, 'description': None,
-          'build_command': 'trick-CP', 'size': 2200000, 'env': None}
         self.config_errors = False
         self.compare_errors = False     # True if comparison errors were found
         self.sims = []                  # list of Sim() instances, filled out from config file
@@ -113,6 +110,9 @@ class TrickWorkflow(WorkflowCommon):
         self.config = self._read_config(self.config_file) # Contains resulting Dict parsed by yaml.load
         self.trick_dir = trick_dir
         self.trick_host_cpu = self.get_trick_host_cpu()
+        # If not found in the config file, these defaults are used
+        self.defaults = {'cpus': 3, 'name': None, 'description': None,
+          'build_command': os.path.join(self.trick_dir,'bin/trick-CP'), 'size': 2200000, 'env': None}
         self._validate_config()
 
     def create_test_suite(self):
```

jdeans289: I like the idea of having TrickOps automatically append the whole path better than what I ended up doing, but this particular snippet of code wouldn't work for me since I have to use 'trick-CP -t' and this only sets the defaults. Maybe either have a flag to use the absolute path, or just do it automatically? I can't think of a case where you would want to use a build command that isn't just arguments to trick-CP

alexlin0: All calls to all trick utilties, trick-CP, trick-ICG, trick-dp, etc.  should be made with the full path.  Otherwise you are vulnerable to users having the wrong environment and trick-CP can be called from the wrong Trick directory.

alexlin0: This applies to all scripts, makefiles, system calls in code, everything.

ddj116: Agreed. I can only think of `trick-CP <args>` as the use case for specifying `build_command`, so I like your suggestion of just tacking on the full absolute path automatically. I probably should have done it this way to begin with, oversight on my part.

alexlin0: It's a lesson I've learned many times. 🤕

share/trick/trickops/TrickWorkflow.py
```diff
@@ -390,11 +391,22 @@ def type_expected(var, expected_type, fatal=False, extra_msg=''):
...
+
+          # Add the full path to the build command
+          self.config[s]['build_command'] = os.path.join(self.trick_dir, "bin", self.config[s]['build_command'])
+
```
jdeans289, line 403: I attach the full path and format the binary string here while the configuration is being processed, which is maybe not the best way to do it. The main issue is that if a user creates a run or sim manually in their script, neither of those things will be applied. This is a problem related to the class structure of TrickOps - the Sim and Run class do not have access to the trick directory or host cpu (although that could be obtained if it had the trick_directory). I couldn't figure out a way around that without either using globals or changing the interface which I am hesitant to do. 

Thoughts?

jdeans289: Or maybe it would be better to take the build_cmd option away all together and have build_args or something instead? That doesn't solve the data access issue, but it would make the user interface a little cleaner I think.

test_sims.yml
```diff
@@ -0,0 +1,235 @@
+
+
```
jdeans289, line 2: All of the test sims, both from the test/ directory and the trick_sims/ directory, are in the yml file at the root of the project. Do yall like that? I could also possibly have 2 separate yml files that live in the test/ and trick_sims/ directories with all their respective sims

trickops.py
```diff
@@ -0,0 +1,38 @@
...
+      run_jobs.remove(first_phase_job)
+
+      builds_status = self.execute_jobs(build_jobs, max_concurrent=8, header='Executing all sim builds.')
+      first_phase_run_status = self.execute_jobs([first_phase_job], max_concurrent=8, header="Executing required first phase runs.")
+      runs_status   = self.execute_jobs(run_jobs,   max_concurrent=8, header='Executing all sim runs.')
+      comparison_result = self.compare()
+      analysis_status = self.execute_jobs(analysis_jobs, max_concurrent=8, header='Executing all analysis.')
+
+      self.report()           # Print Verbose report
+      self.status_summary()   # Print a Succinct summary
```
jdeans289, line 34: The summary gives a nice overview of what passed and failed, and the report will still be generated as usual, but the actual logs are dumped in a file and lost to CI. Maybe I could have this script dump failing logs to stdout, or I could try to find some way to upload them through actions?

share/trick/trickops/TrickWorkflow.py
ddj116: Agree that `build_args` probably makes more sense if we've determined there's no use for `build_cmd` beyond specifying `trick-CP <args>`. I'm all for cleaning up the interface and making the class architecture more useful. TrickOps is very young and this type of critical review from other Trick users / developers if exactly what the code needs.  Right now my teams are the only ones using TrickOps, so we can easily adapt to any framework changes that come out of this review process. All that to say, let's change TrickOps to make it better, don't worry about affecting downstream uesrs, I represent all of them!

test_sims.yml
ddj116: I like the sims all defined in one spot, although separating them might make sense if there's a difference in how you want to test them in CI.  Y'all know more about the Trick-specific workflow and what makes sense so my opinion isn't super relevant 😝

trickops.py
jdeans289: Yeah, this is more of a convenience thought for our own CI with all of it's special problems. You're right that it would be better practice to upload the logfiles, but we've had issues with uploading and finding our reports as is and I don't want all those error reports to evaporate into thin air. I'm not suggesting adding this to TrickOps (I have it implemented locally through the user side script), but I think it's a reasonable workaround until we fix our Github Actions setup.

spfennell (changes requested):

jmpenn (approved):
This looks great! Looks clean and well organized.

spfennell (approved):

Issue #1384, opened 2022-11-04
Replace sprintf() with snprintf() due to security concerns.


jmpenn:
On MacOS, Xcode: "sprintf is deprecated: This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead."

jmpenn:
Hmm SWIG version 4.0.2 generates wrapper code, ( in our case sim_services_wrap.cpp) which contains a call to sprintf(). So, we can't eliminate this/these calls to sprintf() until SWIG fixes this.

PR #1392, opened 2022-11-11
Replace sprintf with snprintf in all of Trick source. #1384

jmpenn:
On MacOS, Xcode: "sprintf is deprecated: This function is provided for compatibility reasons only. Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead."

**NOTE:**
SWIG version 4.0.2 generates wrapper code ( in our case sim_services_wrap.cpp ) which contains a call to ```sprintf()```. So, we can't eliminate this/these calls to ```sprintf()``` until SWIG fixes this.


trick_source/sim_services/CheckPointAgent/ClassicCheckPointerAgent.cpp
```diff
@@ -263,7 +262,7 @@ static int getCompositeSubReference(
         if (referenceOffset == 0) {
             reference_name[0] = '\0' ;
         } else if (referenceOffset > 0) {
-            sprintf(reference_name, " + %ld" , referenceOffset);
+            snprintf(reference_name, (size_t)256, " + %ld" , referenceOffset);
```
spfennell, line 265:
```suggestion
            snprintf(reference_name, 256u, " + %ld" , referenceOffset);
```

spfennell: Pretty sure we could use an unsigned literal instead of casting to size_t. We could use u or ull for all of the size_t casts in this file.

spfennell: Not sure why we need the cast at all. Let me know, I'm curious!

jdeans289 (changes requested):
These changes all look good to me, but unfortunately the bit that was failing on my Mac is in Swig generated code so it can't be addressed at this point, but I'm still getting the failing test locally. The test that's failing is test/SIM_target_specific_variables, and it's failing because it adds `-Werror` to `TRICK_CXX_FLAGS` in its S_overrides.mk. I'm not sure what can be done about that at this point besides just removing the Werror (maybe just for Mac?) 

We should probably document the issue with Swig and snprintf somewhere, since it could cause people's sims to fail mysteriously on Macs if they try to treat warnings as errors.

Side note - I am annoyed that Github won't let me comment on files that haven't been changed, unless there's some way to do it that I can't find.

trick_sims/SIM_sun/models/Sun/src/sun_pred_fast_display.c
```diff
@@ -55,17 +55,15 @@ int sun_pred_fast_display(
 	char message[500];
 
 	message[0] = '\0';
-	sprintf(tmp_s,"%ls %0d:%0d:%02.f",S->label_UTC, S->utc.hour , S->utc.min , S->utc.sec );
+	snprintf(tmp_s, sizeof(tmp_s), "%ls %0d:%0d:%02.f",S->label_UTC, S->utc.hour , S->utc.min , S->utc.sec );
 	strcat(message,tmp_s);
```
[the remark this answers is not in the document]
jmpenn: >> but while we are here ...

I understand. I saw a lot of other things as well, but decided to "only chase one squirrel at a time". After I bag the "sprintf squirrel",  I'll go after the others. ;)

jmpenn:
Jackie, Perhaps we could take it out but leave a comment about it  in the  S_overrides.mk. 

>> We should probably document the issue with Swig and snprintf somewhere.

Agreed, perhaps in a document entitled "Known Issues We Can't Currently Do Anything About."

trick_source/sim_services/CheckPointAgent/ClassicCheckPointerAgent.cpp
jmpenn: The cast to size_t seems clearer, more explicit to me. No?

spfennell: In this case it really doesn't matter because we are casting a numerical literal to a size_t. However, I think we should try not to add new C-style casts to our source code because they:

- Can silence useful compiler errors and warnings about type conversions
- C-style casts can be interpreted in many different ways: https://en.cppreference.com/w/cpp/language/explicit_cast
- Instead, we can choose which kind of cast by explicitly using the C++ provided cast types static_cast, reinterpret_cast, dynamic_cast, const_cast
- c++ casts are more easily searchable 

Also in this specific case the implicit cast from ull to size_t accomplishes the goal. If for some reason the code was refactored later to use a double literal or something the size_t cast would truncate the value and possibly eliminate a compiler error or warning. There is a proposal to add size_t literals to C++23 to solve this very problem (uz and z).

https://www.cprogramming.com/reference/typecasting/cstyle.html#:~:text=C%2Dstyle%20typecasts%20are%20available,integer%20into%20a%20pointer%20type).
another link
ctrl+f static cast in this link:
https://www.stroustrup.com/bs_faq2.html
another reference:
https://isocpp.org/wiki/faq/coding-standards#pointer-casts
 https://en.cppreference.com/w/cpp/language/explicit_cast

jdeans289:
> 

This looks good to me!

jdeans289 (commented):
:)

PR #1398, opened 2022-11-18
Add test sim for interaction between data recording and checkpointing

jdeans289:
Clean version of #1378


jdeans289:
It seems like one of the tests has revealed a race condition in checkpointing, which causes a segmentation fault when the checkpoint is reloaded. It appears that sometimes the variable name and/or variable alias in the data recording group does not get checkpointed correctly.

char* trick_anon_local_8[1];
char* trick_anon_local_9[1];
char trick_anon_local_10[23];
char trick_anon_local_11[1];
...
In a correct checkpoint:

trick_anon_local_8 = 
    {&trick_anon_local_10[0]};

trick_anon_local_9 = 
    {&trick_anon_local_11[0]};

In a buggy checkpoint:

trick_anon_local_8 = 
    {&trick_anon_local_10[0]};

trick_anon_local_9 = "";

For some reason, the variable_alias (sometimes variable_name as well, but less common) is not getting hooked up to the correct reference. This happens in about 1/30 runs, and only on RUN_test6

For now I have RUN_test6 commented out. I've taken a few stabs at debugging this with no progress. This may be worth spinning off into a separate issue, and merging this test without the RUN_test6.

PR #1401, opened 2022-11-21
Refactor workflow distros and workflow python matrix (#1400)

### LINKED ISSUE #1400: Testing every OS with every version of python is too time consuming
Trick devs are spending too much time waiting on tests to pass on a variety of different OSes

The number of Linux distros has been reduced in the main pull_request workflow dispatch
The alternative distros and python 2 tests have been moved to scheduled and/or dispatchable workflows

spfennell:
closes #1400

### DIFF
--- a/.github/workflows/code_coverage.yml
+++ b/.github/workflows/code_coverage.yml
@@ -1,9 +1,14 @@
 name: Code Coverage
 
 on:
-  schedule:
-  - cron:  '0 0 * * 0'
-  workflow_dispatch:
+  push:
+    branches:
+      - master
+    paths-ignore:
+    - 'docs/**'
+    - '.github/workflows/**'
+    - '!.github/workflows/code_coverage.yml'
+  pull_request:
 
 jobs:
   code-coverage:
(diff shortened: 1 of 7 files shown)


spfennell:
@jdeans289 you okay with code coverage being once weekly or on workflow dispatch (you can launch the code coverage test on any branch with the click of a button)?

.github/workflows/code_coverage.yml
```diff
@@ -1,14 +1,9 @@
...
-  push:
-    branches:
-      - master
-    paths-ignore:
-    - 'docs/**'
-    - '.github/workflows/**'
-    - '!.github/workflows/code_coverage.yml'
-  pull_request:
+  schedule:
+  - cron:  '0 0 * * 0'
```
jdeans289, line 5: I think we should leave the code coverage workflow on pull requests and master instead of moving it to a cron job, that way we can see differences on pull requests.

spfennell: Got it, reverted.

PR #1406, opened 2022-11-23
Variable Server Integration Test

jdeans289:
- Update variable server integration test to work and be more thorough
- Add an ability to rerun jobs to Trickops script (user side, not in TrickOps)

This variable server test was written by me awhile ago, but I kept it out of the test suite since it would intermittently fail to connect in CI, which is unrelated to the test. Hopefully, allowing the whole test to be retried will solve the connection issue, but I think the only way to test that is to let it run a few times.


jdeans289:
Looks like this test just fails with the variable server as-is on Linux, even though it should work.

jdeans289:
Closing in favor of #1407

jdeans289:
Reopening this - the issue I was seeing before was related to #1410

trickops.py
```diff
@@ -38,25 +40,65 @@ def run( self ):
...
+      for job in retry_allowed_jobs:
+         run_jobs.remove(job)
+
+
       builds_status = self.execute_jobs(build_jobs, max_concurrent=self.cpus, header='Executing all sim builds.')
       first_phase_run_status = self.execute_jobs(first_phase_jobs, max_concurrent=self.cpus, header="Executing required first phase runs.")
       runs_status   = self.execute_jobs(run_jobs,   max_concurrent=self.cpus, header='Executing all sim runs.')
+
+      # Run the retry_allowed jobs
+      self.execute_jobs(retry_allowed_jobs, max_concurrent=self.cpus, header='Executing retry-allowed runs.')
```
jdeans289, line 56: Sometimes the sockets inexplicably fail to connect in CI and fail the pipeline unnecessarily, so I did a lot of work here to allow retries. I'm not sure that this would even work.

An alternate way to solve the connection problem could be allowing the failed connection exit code to pass the pipeline as well. This integration test is written to return 0 on success, 1 on failure, and 2 on connection failure, so I could try allowing a return code of 2 to pass as well. Allowing multiple return codes isn't natively supported through TrickOps, but it shouldn't be too difficult to hack in.

test/SIM_test_varserv/RUN_test/unit_test.py
```diff
@@ -7,16 +7,18 @@ def main():
...
-	trick_utest.unit_tests.set_test_name( "VariableServerTest" )
+	varServerPort = trick.var_server_get_port()
+	test_output = ( os.getenv("TRICK_HOME") + "/trick_test/SIM_test_varserv.xml" ) 
+	command = 'os.system("./models/test_client/test_client ' + str(varServerPort) + ' --gtest_output=xml:' + test_output + ' &")'
 
-	TRICK_EXPECT_EQ(trick.var_server_get_port(), 40000, "VariableServerTest", "SetPortNumber")
-
-	trick.exec_set_terminate_time(3000.0)
+	# Start the test client after everything has been initialized (hopefully)
+	trick.add_read(1.0, command)
```
jdeans289, line 21: Maybe the reason it failed to connect was a race condition with variable server initialization and this will help?

trickops.py
spfennell: I think the connection failure is a necessary reminder that we should get our socket library to work more reliably

spfennell (approved):

jdeans289:
I thought of some more things that should be tested in here, converting to draft while I work on that

test/SIM_test_varserv/models/test_client/test_client.cpp
```diff
@@ -0,0 +1,509 @@
...
+
+    // Clear out anything else that's been sent
+    // I may need to write something else for this
+    socket << "trick.var_pause()\n";
+    socket.clear_buffered_data();
+
+    // Copy mode 1 (VS_COPY_SCHEDULED) write mode 0 (VS_WRITE_ASYNC)
+    socket << "trick.var_clear()\ntrick.var_set_copy_mode(1)\ntrick.var_add(\"vsx.vst.c\")\ntrick.var_add(\"vsx.vst.d\")\ntrick.var_unpause()\n";
+    socket >> reply;
+
```
jdeans289, line 423: The behavior I'm seeing in this test is that when you use copy mode VS_COPY_SCHEDULED or VS_COPY_TOP_OF_FRAME with write mode VS_WRITE_ASYNC, the first message may have all 0 for data since the first copy hadn't occurred before the first write, and there's no synchronization preventing unstaged data from going out. Is this what we want? Should I preserve this behavior in the refactor or fix it?

test/SIM_test_varserv/models/test_client/test_client.cpp
```diff
@@ -0,0 +1,510 @@
...
+        // 0\t<time> {s}\t
+        // We only care about <time>
+        std::stringstream message_stream(message);
+        std::string token;
+        std::getline(message_stream, token, '\t');
+        std::getline(message_stream, token, ' ');
+        return std::stod(token);
+    };
+
+    // Tail recursion, just for fun
```
spfennell, line 332: No fun allowed! Tail recursion because its good

PR #1407, opened 2022-11-23
Variable Server Refactor

jdeans289:
Rewrite of variable server and integration test


jdeans289:
I should be logging bugs I've found in the original variable server -
- var_byteswap doesn't work
- w_string doesn't work
- Restart of additional listen threads doesn't work 
- (a small amount of) Memory stomping when reading booleans
- Large binary messages don't work
- Relies on undefined behavior in vs_format_ascii
- Various failures to check return codes that can cause the sim to fail in confusing ways

jdeans289:
More bugs that are fixed here - 
- MonteCarlo spawns an extra VariableServerListenThread that can never be closed
- The variable server thread continues listening during restart and can start a thread on a false positive that crashes
- VSThread shutdown was all kinds of messed up
- Random hangs on shutdown

jdeans289:
Bug fixed - additional listen threads were not correctly handled during checkpoint restart

jdeans289:
Bug fixed - checkpoint restart clears unit requests in the variable server session

jdeans289:
Leaving a note for later investigation - there are special varserver commands to send sie files to the client. The S_sie.resource file is generated and used by TrickTV, but I don't see that the other versions of this file are generated or still in use (S_sie_class.xml, S_sie_enum.xml, S_sie_top_level_objects.xml). Is this outdated, or are they used in some case I'm not catching?

Issue #1409, opened 2022-11-29
Simulation works but Trick View not updated in Viewing Real-Time Data with Trick View (TV) Tutorial


Fjolnirr:
Hi,
I was following the trick tutorial and in the [**"Viewing Real-Time Data with Trick View (TV)"**](https://nasa.github.io/trick/tutorial/ATutTrickView) section i am stuck at the ***7th*** step.<br />    Which is ;
**->**_7. On the sim control panel, click the Start button to put the simulation into action. Notice that TV parameter values have changed. When the simulation time reaches 2.0 seconds, the cannon ball position will be [86.17, 30.33]._ <br />
However as you can see in the image below even though my simulation stops at the sec 2.0 There is no update on the Trick View, no value changes or simulation time changes. Am I missing something? Could you please help me?<br />
![bug](https://user-images.githubusercontent.com/40665614/204540012-09c8af9d-6592-433a-9ae5-43adbf84fcf1.PNG)

Edit: Also on run-time cmd prompts this messages; 
```	at java.base/jdk.internal.math.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:1842)
	at java.base/jdk.internal.math.FloatingDecimal.parseDouble(FloatingDecimal.java:110)
	at java.base/java.lang.Double.parseDouble(Double.java:543)
	at trick.common.utils.vs.VSDouble.fromVariableServer(VSDouble.java:15)
	at trick.common.utils.vs.Variable.setValue(Variable.java:129)
	at trick.common.utils.vs.VariableTracker.process(VariableTracker.java:109)
	at java.desktop/javax.swing.SwingWorker$3.run(SwingWorker.java:423)
	at java.desktop/sun.swing.AccumulativeRunnable.run(AccumulativeRunnable.java:112)
	at java.desktop/javax.swing.SwingWorker$DoSubmitAccumulativeRunnable.run(SwingWorker.java:847)
	at java.desktop/sun.swing.AccumulativeRunnable.run(AccumulativeRunnable.java:112)
	at java.desktop/javax.swing.SwingWorker$DoSubmitAccumulativeRunnable.actionPerformed(SwingWorker.java:857)
	at java.desktop/javax.swing.Timer.fireActionPerformed(Timer.java:317)
	at java.desktop/javax.swing.Timer$DoPostEvent.run(Timer.java:249)
	at java.desktop/java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:313)
[...]

PR #1410, opened 2022-11-29
Patch for #1409

jdeans289:
Using the same buffer for the destination and as a parameter for sprintf is undefined behavior. We were doing this previously all over vs_format_ascii without issue:
```
sprintf(value, "%s {%s}", value, ref->units);
```
However, snprintf seems to handle that undefined behavior differently, and was overwriting anything else in value in this specific place. 

This pull request is meant to be a hotfix, but this whole section of the variable server really needs to be refactored.

Link #1409


ddj116:
Thanks for this @jdeans289 -- yesterday I was trying to figure out why the variable server wasn't working on 19.5.0, Scott told me at lunch to stop my investigation because you already had a fix!

PR #1413, opened 2022-12-06
Add ability to set the priority of the socket of the variable server.

stephenmontgomerysmith:


stephenmontgomerysmith:
Oops, I just realized I left a comment in trick_source/trick_utils/comm/src/tc_accept.c.  This should really be a separate issue.  setsockopt doesn't change the value of the_socket, rather the return value should be tested to see if it less than 0.

jdeans289:
Hi Stephen, sorry for the delay. We've been working for awhile on refactoring the variable server code, and now that change has gone into master. I'm going pull branch this PR into a branch off master and integrate your changes manually, since the underlying networking library has changed and everything will need to be rewritten. I will make sure that you still get credit in the commit log and in code. Thank you for your contribution!

PR #1415, opened 2022-12-07
Integrate MonteCarloGenerate capability from EG CML and associated TrickOps enhancements

ddj116:
MonteCarloGenerate: Integrate ramtares CML monte-carlo model into Trick core

* This is a continuation of work from Isaac Reaves' internship in the
  fall of 2022. Original PR: https://github.com/nasa/trick/pull/1389
* Provides alternative to "classic" Trick monte-carlo, but only within
  the run generation mechanism. MonteCarloGenerate (MCG) breaks the
  resource contention problem between master-slave monte-carlo and
  modern HPC systems like PBS and SLURM. See documentation for details
  on how to use MCG.
    
TrickOps: phase, yaml parsing, and [min-max] enhancements
     
* Add new "phase:" mechanism to TrickOps Runs and Builds to support
  project-specific constraints on build and run ordering
  - phase defaults to zero if not specified and must be between -1000
    and 1000 if given.
  - jobs can now optionally be requested by their phase or phase range
  - See trickops/README.md for details
* Add [min-max] notation capability to run: entries and compare: entries
  - [min-max] ranges provide definition of a set of runs using a common
    numbering scheme in the YAML file, greatly reducing YAML file size
    for monte-carlo and other zero-padded run numbering use cases
  - See trickops/README.md for details
* YAML parsing changes
  - Overhaul the logic which verifies YAML files for the expected
    TrickOps format. This is now done in TrickWorkflowYamlVerifier and
    provides much more robust error checking than previous approach
  - .yaml_requirements.yml now provides the required types, ranges, and
    default values as applicable to expected entries in YAML files
  - valgrind: is now an sub-option to run: entries, not its own section
    Users should now list their runs normallly and define their flags in
    in that run's valgrind: subsection
  - parallel_safety is now a per-sim parameter and not global. Users
    should move their global config to the sim layer
  - self.config_errors is now a list of errors. Users should now
    check for empty list when using instead of True/False
* Robustify the get_koviz_report_jobs unit test to work whether koviz
  exists on PATH or not
* Adjust trickops.py to use the new phase and range features
   - Make it more configurable on the command-line via argparse
   - Move SIM_mc_generation tests into test_sims.yml
* Disable var server connection by default in SingleRun if
  TrickWorkflow.quiet is True
* Guard against multiple Job starts in Job class
* Remove SimulationJob inheritance layer since old monte-carlo wasn't
  and never will be supported by TrickOps


ddj116:
I've run into a bit of an issue in digging into the failures associated with Ubuntu-22.  `trickops.py` fails on four comparisons:

```
  FAIL  test/SIM_mc_generation/verif_data/MONTE_RUN_random_uniform/monte_values_all_runs vs. test/SIM_mc_generation/MONTE_RUN_random_uniform/monte_values_all_runs
  FAIL  test/SIM_mc_generation/verif_data/MONTE_RUN_random_uniform/RUN_2/monte_input.py vs. test/SIM_mc_generation/MONTE_RUN_random_uniform/RUN_2/monte_input.py
  FAIL  test/SIM_mc_generation/verif_data/MONTE_RUN_random_uniform/RUN_4/monte_input.py vs. test/SIM_mc_generation/MONTE_RUN_random_uniform/RUN_4/monte_input.py
  FAIL  test/SIM_mc_generation/verif_data/MONTE_RUN_random_uniform/RUN_9/monte_input.py vs. test/SIM_mc_generation/MONTE_RUN_random_uniform/RUN_9/monte_input.py
```
When digging into the details, we find that `std::uniform_int_distribution<int>` is providing different values for different compiler versions.  Here's a summary (the differences between `monte_values_all_runs`) across platforms/compilers:

```
--------------------------------------------------------------------------------------
    VERIF DATA (from gcc 4.8)                     GENERATED DATA (from gcc 11.3)
  run, test.x_uniform, test.x_integer             run, test.x_uniform, test.x_integer
--------------------------------------------------------------------------------------
  0, 66000.76700176572, 47109                  |  0, 66000.76700176572, 47109
  1, 81004.55299703917, 78712                  |  1, 81004.55299703917, 78712
  2, 84161.578104146,   91306                  |  2, 84161.578104146,   91305  <---  less by one
  3, 5338.88635155008,   2546                  |  3, 5338.88635155008,   2546
  4, 81653.07878167802, 28114                  |  4, 81653.07878167802, 28113  <---  less by one
  5, 74245.08441568459, 18377                  |  5, 74245.08441568459, 18377
  6, 7306.365748693855, 82893                  |  6, 7306.365748693855, 82893
  7, 83068.39187988771, 76648                  |  7, 83068.39187988771, 76648
  8, 58483.47941203944,   887                  |  8, 58483.47941203944,   887
  9, 1930.026637775263, 96421                  |  9, 1930.026637775263, 96420  <---  less by one
```
In order to confirm that this is indeed coming from the `std` methods and not our model itself, I was able to replicate the same behavior in a small standalone `c++` program:

```c++
#include <random>
#include <iostream>
int main()
{
// Data associated with test/SIM_mc_generation/RUN_random_uniform/input.py, line that looks like...
// mc_var = trick.MonteCarloVariableRandomUniformInt( "test.x_integer", 77001, 100, 100000)
    unsigned int seed = 77001;
    double lower_bound = 100.0;
    double upper_bound = 100000.0;
    std::mt19937 random_generator(seed);
    std::uniform_int_distribution<int> distribution(lower_bound, upper_bound);
    for (int n=0; n<10; ++n) {
      std::cout << "value for lower bound: " << lower_bound << ", upper_bound: "
        << upper_bound << ", seed: " << seed << " is "  << distribution(random_generator) << std::endl;
    }
}
```
The output of this program on CentOS 7.10 (`gcc` 4.8.5) vs. Ubuntu-22 (`gcc` 11.3.0) confirms my theory, but also reveals a bigger scope in the problem when you consider these tests running on Mac OSX. Here's the output from all three platforms The `*` indicates values that are different from the `verif_data` values, a.k.a the "correct data" as it comes from CML:

```
---------------------------------------------------------------------------------
10 random ints using std::mt19937 and std::uniform_int_distribution<int>  with 
  lower_bound=100, upper_bound=100000, seed = 77001
  * indicates a different value from the expected verif_data from CML
---------------------------------------------------------------------------------
int CentOS 7.10 (gcc 4.8.5)  Ubuntu 22 (gcc 11.3.0)  Mac OSX (Apple clang 14.0.0)
0        47109                      47109                       42597*
1        78712                      78712                       29615*
2        91306                      91305*                      71364*
3        2546                       2546                        93319*
4        28114                      28113*                      10651*
5        18377                      18377                       79104*
6        82893                      82893                       21531*
7        76648                      76648                       19116*
8        887                        887                         96817*
9        96421                      96420*                      50007*
```
@jdeans289 @jmpenn - we should discuss how to move forward on this with respect to `trickops.py`. We could attempt to provide expected data for each platform, and then have `trickops.py` check the platform to know which files to compare, or we could disable the comparisons for certain platforms/compilers, or we could disable the comparisons for all of `SIM_mc_generation` entirely.  The TrickOps YAML file has no concept of "this `compare:` section only applies to this platform", but we could still perform that logic in `trickops.py` as a "project specific need".  

Will swing by to chat about it.

trick_source/sim_services/MonteCarloGeneration/mc_variable_random_normal.cc
```diff
@@ -0,0 +1,267 @@
...
+{
+  truncated_low = false;
+  truncated_high = false;
+}
+
+/*****************************************************************************
+ * is_near_equal
+ * Purpose: Utility function providing double==double equality checks
+ * **************************************************************************/
+bool MonteCarloVariableRandomNormal::is_near_equal( double val1, double val2)
```
ddj116, line 237: FYI @jmpenn this is the function in question, originally written by Gary to prevent FPEs from terminating the sim during the check.  I slightly altered the `fe` system calls from the original so that we are using functions that exist  on both Mac OSX and linux.

If you have questions on the precise types of FPEs that could be thrown, Gary might have some insight into that. Yesterday in the Tricklab John had mentioned that if only underflows are of concern, most OSs automatically translate those to zero. Ultimately we just want a safe "are these two doubles equivalent?" check somewhere in the Trick framework. If y'all have a better function for us to use, we can easily change the model to point to that, just let us know.

share/trick/trickops/TrickWorkflow.py
```diff
@@ -100,20 +177,112 @@ def __init__(self, project_top_level, log_dir, trick_dir, config_file, cpus=3, q
...
+                 % (s, self.config[s]['path'], self.config_file), 'DARK_RED')
+            continue
+          if self.config[s]['path'] in all_sim_paths:
+            cprint("ERROR: %s's 'path' is not unique in %s. Continuing but ignoring this sim."
+                % (s, self.config_file), 'DARK_RED')
+            continue
+          # Add the CPU format string (won't fail if there is no CPU placeholder)
+          self.config[s]['binary'] = self.config[s]['binary'].format(cpu=self.trick_host_cpu)
+          # Add the full path to the build command
+          self.config[s]['build_command'] = os.path.join(self.trick_dir, "bin", self.config[s]['build_command'])
```
ddj116, line 224: Hey @jdeans289 I just came across a potential use of `build_command` that this isn't compatible with. One of my gateway users is trying to do this:
Screen Shot 2023-01-20 at 9 28 01 AM
I think we can pretty easily support this by doing a `str.replace('trick-CP', os.path.join(self.trick_dir, "bin/trick-CP"))`, but I guess the question I have is whether or not you think this should be allowed. A few weeks ago we asked out loud if there was any use case beyond flags passed to `trick-CP`. Well here's one - setting an environment variable that should only apply to one sim's build.  What do you think?

ddj116: I've replace `build_command` with `build_args` per discussion today. Simpler and less prone to error 👍

share/trick/trickops/TrickWorkflow.py
```diff
@@ -1380,20 +1474,25 @@ def connect():
                     return
                 except socket.timeout:
                     pass
+                # If a SingleRun job terminates before the thread can connect to the
+                # variable server, Trick's variable_server module throws an IOError
+                # with the message: The remote endpoint has closed the connection
+                except IOError as e:
```
ddj116, line 1480: FYI @jdeans289 this resolves the `IOError("The remote endpoint has closed the connection")` problem we see when the variable server tries to connect to a sim that has already terminated.  In case you're curious, the easiest way to replicate this is by instrumenting one of the test sims (I used `SIM_ball_L1`) with this type of "long-initialization, short runtime" setup:

```python
trick.exec_set_terminate_time(1.0)
trick.add_read(0.0, "import time; time.sleep(10)")
```
This gives TrickOps enough time to attempt a connection but not enough time for the handshake to complete.

ddj116:
OK I've replicated the MacOS GitHub Actions hang.  It's coming from the execution of generated runs, replicated as follows on this PR's source branch:

1. Build `test/SIM_mc_generation` (suggest modifying `S_overrides.mk` to add `-O0` to `TRICK_CXXFLAGS`)
2. Generate the nominal runs set by running: `./S_main_Darwin_21.exe RUN_nominal/input_a.py`
3. Execute one of the generated runs via: `./S_main_Darwin_21.exe MONTE_RUN_nominal/RUN_000/monte_input_a.py`

(3) hangs for me every time. If you build trick with `-g -O0`, we get an interesting stack trace at the hang using `lldb`:

```gdb
(lldb) thread backtrace all
  thread #1, queue = 'com.apple.main-thread'
    frame #0: 0x00007ff80a13dbd2 libsystem_kernel.dylib`__psynch_mutexwait + 10
    frame #1: 0x00007ff80a175e7e libsystem_pthread.dylib`_pthread_mutex_firstfit_lock_wait + 76
    frame #2: 0x00007ff80a173cbb libsystem_pthread.dylib`_pthread_mutex_firstfit_lock_slow + 205
    frame #3: 0x00000001000eb55b S_main_Darwin_21.exe`Trick::DataRecordDispatcher::shutdown(this=0x000000010085a050) at DataRecordDispatcher.cpp:247:9
    frame #4: 0x0000000100005db2 S_main_Darwin_21.exe`DataRecordDispatcherSimObject::call_function(this=0x0000000100859fc8, curr_job=0x000000010180a2a0) at S_source.cpp:595:17
    frame #5: 0x00000001001e219b S_main_Darwin_21.exe`Trick::JobData::call(this=0x000000010180a2a0) at JobData.cpp:182:26
    frame #6: 0x00000001001346ea S_main_Darwin_21.exe`Trick::Executive::shutdown(this=0x0000000100857690) at Executive_shutdown.cpp:67:27
    frame #7: 0x00000001002c3241 S_main_Darwin_21.exe`master(nargs=2, args=0x00007ff7bfeff858) at master.cpp:48:17
    frame #8: 0x000000010000b3d2 S_main_Darwin_21.exe`main(nargs=2, args=0x00007ff7bfeff858) at S_source.cpp:1336:12
    frame #9: 0x000000010142d52e dyld`start + 462
  thread #2
    frame #0: 0x00007ff80a143d5a libsystem_kernel.dylib`__select + 10
    frame #1: 0x00000001001f963c S_main_Darwin_21.exe`Trick::VariableServerListenThread::thread_body(this=0x0000000100859490) at VariableServerListenThread.cpp:174:9
    frame #2: 0x00000001001eb1b0 S_main_Darwin_21.exe`Trick::ThreadBase::thread_helper(context=0x0000000100859490) at ThreadBase.cpp:321:16
    frame #3: 0x00007ff80a1784e1 libsystem_pthread.dylib`_pthread_start + 125
    frame #4: 0x00007ff80a173f6b libsystem_pthread.dylib`thread_start + 15
* thread #3, stop reason = EXC_BAD_ACCESS (code=1, address=0x43424100)
    frame #0: 0x00007ff80a18b6b2 libsystem_platform.dylib`_platform_strlen + 18
    frame #1: 0x00007ff80a044eaa libsystem_c.dylib`__vfprintf + 5000
    frame #2: 0x00007ff80a076868 libsystem_c.dylib`_vsnprintf + 249
    frame #3: 0x00007ff80a0512d0 libsystem_c.dylib`vsnprintf_l + 41
    frame #4: 0x00007ff80a051389 libsystem_c.dylib`snprintf + 156
  * frame #5: 0x00000001000e74ee S_main_Darwin_21.exe`Trick::DRAscii::copy_data_ascii_item(this=0x000000010281da00, DI=0x0000600002601740, item_num=0, buf="") at DRAscii.cpp:201:13
    frame #6: 0x00000001000e7368 S_main_Darwin_21.exe`Trick::DRAscii::format_specific_write_data(this=0x000000010281da00, writer_offset=0) at DRAscii.cpp:127:9
    frame #7: 0x00000001000f1e75 S_main_Darwin_21.exe`Trick::DataRecordGroup::write_data(this=0x000000010281da00, must_write=true) at DataRecordGroup.cpp:707:36
    frame #8: 0x00000001000ea6de S_main_Darwin_21.exe`Trick::DRDWriterThread::thread_body(this=0x000000010085a088) at DataRecordDispatcher.cpp:58:29
    frame #9: 0x00000001001eb1b0 S_main_Darwin_21.exe`Trick::ThreadBase::thread_helper(context=0x000000010085a088) at ThreadBase.cpp:321:16
    frame #10: 0x00007ff80a1784e1 libsystem_pthread.dylib`_pthread_start + 125
    frame #11: 0x00007ff80a173f6b libsystem_pthread.dylib`thread_start + 15
```
I don't quite understand why an `EXC_BAD_ACCESS` exception would cause the binary to hang on Darwin (I'd expect a crash...), but this morning I found out that this coming from the sim data recording two `std::string`s:
```python
    dr_group.add_variable( "test.x_string")
    dr_group.add_variable( "test.x_fixed_value_string")
```
I already talked with @spfennell and @jmpenn about a year ago and learned that Trick doesn't support data recording strings - this is now the 2nd project I've run into where folks are doing this, thinking it is supported. It seems to accidentally work on Linux but Darwin is good about catching the memory faults. I'm going to edit `Data-Record.md` to make it clear that groups should not be doing this. Then I'll modify `SIM_mc_generation`'s data logging and tests so that strings are purged.

ddj116:
On to the next Darwin-only problem on this branch. Many tests in `test/SIM_mc_generation` fail, alternating between two different modes. For example, let's look at `test/SIM_mc_generation`'s `RUN_random_normal_truncate_abs/input.py`. Sometimes it runs and returns 0, but with many `tc_accept` errors:

```
./S_main_Darwin_21.exe RUN_random_normal_truncate_abs/input.py
|L   0|2023/01/27,13:24:40|JSLAL0520093471| |T 0|0.000000| 
SIMULATION TERMINATED IN
  PROCESS: 0
  ROUTINE: mc_master.cc:408
  DIAGNOSTIC: 
Monte-Input files generated in MONTE_RUN_random_normal_truncate_abs
Runs can now be launched using the generated monte-input files and
managed with an external load-management system, such as SLURM.
This simulation is complete.  Exiting.


       SIMULATION START TIME:        0.000
        SIMULATION STOP TIME:        0.000
     SIMULATION ELAPSED TIME:        0.000
        ACTUAL CPU TIME USED:        0.228
       SIMULATION / CPU TIME:        0.000
     INITIALIZATION CPU TIME:        0.000

 ALERT: MessageTCDevice.cpp:80: tc_accept: (ID = 0  tag = <empty>): error accepting the pending socket connection
(repeats 75 more times)
```
Other times, the exact same execution results in `Process terminated by signal SIGSEGV` (exit code 11). The `tc_accept` error never prints. When this happens we get this stacktrace:

```gdb
(lldb) thread backtrace all
  thread #1, queue = 'com.apple.main-thread'
    frame #0: 0x00007ff80a13cd96 libsystem_kernel.dylib`__close_nocancel + 10
    frame #1: 0x00007ff80a0679ef libsystem_c.dylib`fclose + 107
    frame #2: 0x00007ff80a0eca99 libc++.1.dylib`std::__1::basic_filebuf<char, std::__1::char_traits<char> >::close() + 39
    frame #3: 0x00000001001788ad S_main_Darwin_21.exe`std::__1::basic_fstream<char, std::__1::char_traits<char> >::close(this=0x0000000100858a38) at fstream:1739:15
    frame #4: 0x00000001001787d7 S_main_Darwin_21.exe`Trick::MessageFile::~MessageFile(this=0x00000001008589f8) at MessageFile.cpp:64:16
    frame #5: 0x0000000100178915 S_main_Darwin_21.exe`Trick::MessageFile::~MessageFile(this=0x00000001008589f8) at MessageFile.cpp:63:36
    frame #6: 0x0000000100015e5e S_main_Darwin_21.exe`MessageSimObject::~MessageSimObject(this=0x00000001008587f8) at S_source.hh:275:7
    frame #7: 0x0000000100007685 S_main_Darwin_21.exe`MessageSimObject::~MessageSimObject(this=0x00000001008587f8) at S_source.hh:275:7
    frame #8: 0x00007ff80a070de4 libsystem_c.dylib`__cxa_finalize_ranges + 409
    frame #9: 0x00007ff80a070bfe libsystem_c.dylib`exit + 35
    frame #10: 0x00007ff80a184375 libdyld.dylib`dyld4::LibSystemHelpers::exit(int) const + 11
    frame #11: 0x0000000101459558 dyld`start + 504
  thread #2
    frame #0: 0x00007ff80a143d5a libsystem_kernel.dylib`__select + 10
    frame #1: 0x00000001001f963c S_main_Darwin_21.exe`Trick::VariableServerListenThread::thread_body(this=0x0000000100859490) at VariableServerListenThread.cpp:174:9
    frame #2: 0x00000001001eb1b0 S_main_Darwin_21.exe`Trick::ThreadBase::thread_helper(context=0x0000000100859490) at ThreadBase.cpp:321:16
    frame #3: 0x00007ff80a1784e1 libsystem_pthread.dylib`_pthread_start + 125
    frame #4: 0x00007ff80a173f6b libsystem_pthread.dylib`thread_start + 15
* thread #3, stop reason = EXC_BAD_ACCESS (code=EXC_I386_GPFLT)
  * frame #0: 0x0000000100570bf4 S_main_Darwin_21.exe`trick_error_report(error_hndlr=0x0000600002c04180, error_level=TRICK_ERROR_ALERT, file="MessageTCDevice.cpp", line=80, format="tc_accept: %s: error accepting the pending socket connection\n") at trick_error_hndlr.c:624:9
    frame #1: 0x000000010056ada5 S_main_Darwin_21.exe`tc_accept_(listen_device=0x0000000100858cf0, device=0x0000600003710000, file="MessageTCDevice.cpp", line=80) at tc_accept.c:38:9
    frame #2: 0x000000010017f296 S_main_Darwin_21.exe`Trick::MessageTCDeviceListenThread::thread_body(this=0x0000000100858cb0) at MessageTCDevice.cpp:80:22
    frame #3: 0x00000001001eb1b0 S_main_Darwin_21.exe`Trick::ThreadBase::thread_helper(context=0x0000000100858cb0) at ThreadBase.cpp:321:16
    frame #4: 0x00007ff80a1784e1 libsystem_pthread.dylib`_pthread_start + 125
    frame #5: 0x00007ff80a173f6b libsystem_pthread.dylib`thread_start + 15
```
@jdeans289 @jmpenn do y'all have any idea what might be causing an exception in `trick_error_report` or how to proceed in diagnosing the issue? What might cause the `tc_accept` errors?

trick_source/sim_services/MonteCarloGeneration/mc_variable_random_normal.cc
jmpenn: I think the reason that FP_underflows are being being generated is on line 256
For very tiny numbers it is generating FP_subnormals which result in FP_overflow exceptions.

From : https://www.gnu.org/software/libc/manual/html_node/FP-Exceptions.html
```
 ‘Underflow’
 The underflow exception is raised when an intermediate result
 is too small to be calculated accurately, or if the operation’s
 result rounded to the destination precision is too small to be normalized.

When no trap is installed for the underflow exception, underflow is
signaled (via the underflow flag) only when both tininess and loss
of accuracy have been detected. If no trap handler is installed the
operation continues with an imprecise small value, or zero if the
destination precision cannot hold the small exact result.
```

Some OS'es have handlers installed that set the result to 0.0. Others don't. 

For a general Trick "close_enough" function, I'd prefer to go with something that avoids FP_subnormals altogether. Below is one I wrote that avoid FP_subnormals. Note that it doesn't use subtraction to do the test : ```if ((fmin(A,B) + tolerance) >= fmax(A,B)) {``` . This is because even if the arguments are both FP_NORMAL, a subtraction can create an FP_SUBNORMAL. Example: 1.5*DBL_MIN - DBL_MIN.


```c
#include <float.h>
#include <math.h>
#include <stdio.h>

int close_enough( double A, double B, double tolerance) {

    // If an arg is a FP_SUBNORMAL that is: less than DBL_MIN (2.22507e-308)
    // then set it to zero.
    if ( fpclassify(A) == FP_SUBNORMAL ) { A = 0.0; }
    if ( fpclassify(B) == FP_SUBNORMAL ) { B = 0.0; }

    // If A and B are identical, then they're close_enough.
    if (A==B) { return true; }

    // The tolerance must be an FP_NORMAL. Nothing else makes sense.
    // Not even FP_SUBNORMAL or FP_ZERO, given the previous tests.
    // Also, a negative tolerance is silly.
    if (( fpclassify(tolerance) != FP_NORMAL ) || (tolerance < 0.0)) { return false; }

    // If the tolerance were less than the gap around the arg with
    // largest magnitude, then the args couldn't possibly be close enough.
    // So, the tolerance must be greater than or equal to the larger of
    // the gaps around A and B.
    if ( tolerance >= DBL_EPSILON * fmax( fabs(A), fabs(B))) {

        // We want to avoid using subtraction in determining
        // whether A and B are within tolerance of each other.
        // The reason is that it might generate a FP_SUBNORMAL result,
        // which would raise an FP_underflow exception. For example:
        // If A = 1.5*DBL_MIN, and B = DBL_MIN, the difference is an
        // FP_SUBNORMAL value. When FP_SUBNORMAL values are generated,
        // so are FP_underflows. By insisting that our args and tolerance
        // are not FP_SUBNORMAL, and by only using addition and comparison we
        // can avoid FP_SUBNORMAL results.
        if ((fmin(A,B) + tolerance) >= fmax(A,B)) {
            return true ; // then they're close enough.
        } else {
            return false ;
        }
    } else {
        return false;
    }
}
```

ddj116:
Note to self, CI testing on hold for this branch until #1448 is finalized, since this branch will fail testing without it. Will rebase with `origin/master` and then give Actions a shot then.

share/trick/sim_objects/default_trick_sys.sm
```diff
@@ -175,7 +175,7 @@ class MonteCarloSimObject : public Trick::SimObject {
             exec_register_scheduler(&mc) ;
 
             {TRK} P0 ("default_data")   mc.process_sim_args() ;
-            {TRK} P0 ("initialization") mc.execute_monte() ;
+            {TRK} P1 ("initialization") mc.execute_monte() ;
```
ddj116, line 178: FYI @jdeans289 @jmpenn I had to increase the phase by one here `P1` so that the new method `MonteCarloGenerate.sm:41` executes first at `P0`. Without this change, a user mis-configuring a SIM for both old and new MC never hits the error checking in `MonteCarloGenerate.sm:43` if `trick_mc.mc.set_dry_run(True)`. Let me know if you have objection to this or another way to get the same behavior.

ddj116:
Looks like MacOS got a segfault on `Failing job:  Run test/SIM_checkpoint_data_recording RUN_test4/unit_test.py`, but this same job passe without issue on my last commit with essentially no changes.  I know y'all mentioned sometimes having trouble with the Mac jobs, should I just hit retry?

jdeans289:
Yeah, I went ahead and reran it, I should probably remove that test until we can fix it. It looks like the rockylinux was hanging, so I reran that one as well. I'm not sure if that's because of the recent threading changes or if there's something else going on.

Issue #478, opened 2017-09-21
Create parent directories with the -O command line option


dbankieris:
`-O` allows the user to specify an output directory. Annoyingly, Trick does not create parent directories as necessary. Granted, there is no `-p` option for the C version of the function. Time to get creative!

alexlin0:
You could use the [create_directories](http://en.cppreference.com/w/cpp/filesystem/create_directory) function in C++17.

dbankieris:
Seems reasonable to me that Trick 17 should require C++ 17!

sharmeye:
Cleaning up issues, reopen if closed in error.

PR #1420, opened 2022-12-16
Allow for read only S_sie.resource

jdeans289:
Move S_sie.resource runtime generation to separate file in output dir
Make output dir create the whole path ( #478 )


jmpenn (commented):
Suggest removing commented-out code in create_path_test.cpp

PR #1432, opened 2023-01-19
Handle empty PYTHON_MODULE in header comment

### LINKED ISSUE #1431: Empty PYTHON_MODULE() statement silently kills the sim
If an empty PYTHON_MODULE() statement is left in a trick header comment, the sim will compile successfully but fail at runtime with mysterious error messages.

Example of errors: 
```
  File "<string>", line 8, in <module>
  File "/Users/jndeans/trick/test/SIM_python_namespace/trick.zip/trick/__init__.py", line 3, in <module>
ModuleNotFoundError: No module named 'm37db6fb550df75f8906f3e6397b65402'
```

The header comment processing script should be robust enough to handle an empty PYTHON_MODULE statement in a way that is informative and intuitive to the user. It should either output an error message and fail the compile, or completely ignore the statement and compile as if it isn't there.

1. An empty PYTHON_MODULE() statement either a) no longer causes this error or b) fails the compile with a message that makes the problem obvious.
2. This behavior is tested and documented.

jdeans289:
closes #1431

### DIFF
--- a/test/SIM_python_namespace/models/TrickFood.hh
+++ b/test/SIM_python_namespace/models/TrickFood.hh
@@ -2,7 +2,7 @@
 @file
 
 @verbatim
-PURPOSE: (Empty python module won't )
+PURPOSE: (Test that an empty PYTHON_MODULE won't cause errors or have any effect.)
 PYTHON_MODULE: ()
 @endverbatim
 *******************************************************************************/
@@ -28,5 +28,5 @@ class Food {
 };
 
 
-#endif /* _BALL_HH_ */
+#endif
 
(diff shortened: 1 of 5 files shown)


test/SIM_python_namespace/models/TrickFood.hh
```diff
@@ -0,0 +1,32 @@
+/**
+@file
+
+@verbatim
+PURPOSE: (Empty python module won't )
```
jmpenn, line 5: Empty python module won't what?  

Empty python module will be ignored?

jmpenn (commented):
Suggest fixing the purpose comment in TrickFood.hh and perhaps the "blah blah blah" comment. Otherwise looks like a straight forward fix.

PR #1434, opened 2023-01-20
Add option to terminate sim on python event error

### LINKED ISSUE #693: Trick events that produce a traceback should terminate the sim
Currently, any trick event that produces a traceback of any kind does not terminate the simulation but does stop the execution of the python event block.  Consider this example, where I define 3 new trick events in `trick_sims/SIM_cannon_eulercromer`:

```
trick.add_read(0.0, """
print "First"
dyn.cannon.acc[0] = trick.attach_units( "d",1.0)
print "End First"
""")
trick.add_read(0.0, """
print "Second"
dyn.cannon.acc = 2.0
print "End Second"
""")
trick.add_read(0.0, """
print "Third"
dyn.cannon.acc[1] = [2.0, 3.0]
print "End Third"
""")

```

Here's the output for these three `add_read` statements:

```
First
Units specifications have different number of units_primitives.
|L   3|2018/11/05,15:00:47|0f1df6e90723| |T 0|0.000000| Memory Manager ERROR: Can't convert "d" to "M/s2".
  File "<string>", line 3, in <module>
AttributeError: Units Conversion error in method swig_ref::__setitem__
Second
  File "<string>", line 3, in <module>
TypeError: Right hand side could not be converted to proper array type
Third
Cannot assign list to a scalar value
  File "<string>", line 3, in <module>
IndexError: Error in method swig_ref::__setitem__
... clipped ...
SIMULATION TERMINATED IN
  PROCESS: 0
  ROUTINE: Executive_loop_single_thread.cpp:93
  DIAGNOSTIC: Reached termination time

       SIMULATION START TIME:        0.000
        SIMULATION STOP TIME:        5.200
     SIMULATION ELAPSED TIME:        5.200
        ACTUAL CPU TIME USED:        0.036
       SIMULATION / CPU TIME:      146.281
     INITIALIZATION CPU TIME:        0.028
~/trick/trick_sims/SIM_cannon_eulercromer > echo $?
0
```

Notice that none of the `End *` prints are executed and the sim returns zero, indicating success.  As a trick user, this seems very intuitive.  Since the sim doesn't terminate and just continues on it's merry way as if the `add_read` logic was never executed in the first place, if the user doesn't see the traceback, they will never know there was a problem.  This is even more of an issue for monte-carlo users, where the output typically whizzes by quite quickly.

One workaround is to define `try/except/terminate` logic around every Trick event that the user creates, but this is painful and prone to error.  Are there legitimate use cases where the user would prefer the current behavior (non-fatal failure and ignore the remaining block)?  It seems like the "fail sim on event failure" would be much more useful to 99% of users, and that is what I'm requesting here.  Or, at least an option to turn it on.

### LINKED ISSUE #1433: Add option to terminate a sim on a Python error in Trick events
The input processor is used to run Trick events or add_read statements that take place at any time during a sim's execution, but unlike the initial input file processing, any errors raised by the Python code will not affect the running sim. This can result in silent python failures, especially in CI pipelines. Add an option to terminate the sim if a Python error is detected at any point during the run.

1. The above feature is implemented, tested, and documented.

jdeans289:
closes #1433 
closes #693 

Adds a command to terminate the sim on event processing failure:
```python
trick.terminate_on_event_parse_error(<True|False>)
```
Tests and documentation update included.

### DIFF
--- a/docs/documentation/running_a_simulation/Input-File.md
+++ b/docs/documentation/running_a_simulation/Input-File.md
@@ -449,6 +449,9 @@ For information on how Trick processes events during runtime, see [Event Process
 
 # Add the event to the input processor's list of events (it will be processed at top of frame before scheduled jobs)
 trick.add_event(<event name>)
+
+# Tell trick whether to terminate the sim if an error occurs while parsing Python code. Defaults to False
+trick.terminate_on_event_parse_error(<True|False>)
 ```
 
 #### Advanced Event (Malfunction) Usage
--- a/include/trick/IPPythonEvent.hh
+++ b/include/trick/IPPythonEvent.hh
@@ -185,6 +185,15 @@ namespace Trick {
             static void set_event_info_msg_off() ;
 
             /**
+             @brief @userdesc Command to set whether the sim should check error codes from
+             Python parsing and terminate if an error is detected. Set to false by default
+             @par Python Usage:
+             @code trick.terminate_on_event_parse_error(True|False) @endcode
+             @return always 0
+            */
+            static void terminate_on_event_parse_error(bool on_off);
+
+            /**
              @brief called by the event manager when the event is loaded from a checkpoint
             */
             virtual void restart() ;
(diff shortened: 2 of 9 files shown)


jmpenn (approved):
Looks like a good, clean change! And it's documented! :)

PR #1436, opened 2023-01-23
Update convert_swig

mrjaperez:
Allow users to specify SWIG interface extensions not supported by Trick. Will broaden the python input processor support.

The current usecase for this is the following:
The project C++ source code must not contain any SWIG templating (using #ifdef SWIG)
The SWIG template extensions must live in a separate file.
The project C++ source requires additional SWIG templating (not supported by Trick) to interface with set std::array and nested classes via the input processor, to name a few.

### DIFF
--- a/libexec/trick/convert_swig
+++ b/libexec/trick/convert_swig
@@ -330,6 +330,16 @@ sub process_file() {
                 $class_typemap_printed{$c} = 1 ;
             }
         }
+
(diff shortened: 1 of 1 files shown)


jdeans289:
Thank you for your contribution, but we are concerned about the permissiveness of the implementation. We'll consider implementing similar functionality in the future.

PR #1437, opened 2023-01-24
Remove unnecessary return; from last line of void functions

thnkslprpt:
Checked through the code for unnecessary `return;` statements on the last line of `void` functions, which are not required.

Improves code consistency (most void functions in trick do not include a `return;` on the final line).

No impact on code logic.


jmpenn:
This pull request appears to add nothing of value.

thnkslprpt:
> This pull request appears to add nothing of value.

The value is improved consistency in the code base, and removing unnecessary (redundant) lines of code.

Anyway, it's your choice.
Best of luck.

PR #1441, opened 2023-01-28
Add check and warning for unsupported types in data recording

### LINKED ISSUE #1440: Add a warning when trying to data record unsupported types
Currently, data recording with a std::string or char * is not implemented correctly and will cause memory problems. Trick should check a data recording variable and not add it if it is a type that does not work, and print a warning to the user.

1. Trying to data record a string type no longer causes memory problems
2. A warning is printed when a user tries to add an unsupported type to data recording
3. This behavior is tested and documented

jdeans289:
closes #1440


test/SIM_test_dr/models/dr/src/DR_default_data.cpp
```diff
@@ -25,7 +25,8 @@ int DRTypes::init() {
...
-
+	o = "3 May. Bistritz.—Left Munich at 8:35 P. M., on 1st May, arriving at Vienna early next morning; should have arrived at 6:46, but train was an hour late. Buda-Pesth seems a wonderful place, from the glimpse which I got of it from the train and the little I could walk through the streets. I feared to go very far from the station, as we had arrived late and would start as near the correct time as possible. The impression I had was that we were leaving the West and entering the East; the most western of splendid bridges over the Danube, which is here of noble width and depth, took us among the traditions of Turkish rule.";
+	p = "We left in pretty good time, and came after nightfall to Klausenburgh. Here I stopped for the night at the Hotel Royale. I had for dinner, or rather supper, a chicken done up some way with red pepper, which was very good but thirsty. (Mem., get recipe for Mina.) I asked the waiter, and he said it was called “paprika hendl,” and that, as it was a national dish, I should be able to get it anywhere along the Carpathians. I found my smattering of German very useful here; indeed, I don't know how I should be able to get on without it.";
```
ddj116, line 29: 🧛

jmpenn (approved):
Good change.

PR #1443, opened 2023-01-30
Initial commit of compareFloatingPoint utils.

jmpenn:
This adds utility functions to compare floating point numbers to determine whether they are within a tolerance of each other.
They are designed to not cause FP_underflows regardless of whether the OS handles underflow or not.


trick_source/trick_utils/compareFloatingPoint/test/.gitignore
```diff
@@ -0,0 +1,2 @@
+*.o
+Interpolator_unittest
```
jdeans289, line 2: I'm thinking this should be dbl_is_near_unittest and flt_is_near_unittest? Or *_unittest

trick_source/trick_utils/compareFloatingPoint/README.md
```diff
@@ -0,0 +1,31 @@
...
+## ```Trick::dbl_is_near```
+```c
+bool Trick::dbl_is_near( double A, double B, double tolerance);
+```
+
+This function compares the values of ```double A``` and ```double B``` to determine whether they are within tolerance of each other. If they are, then the function returns ```true```, otherwise it returns ```false```.
+
+The design of ```Trick::dbl_is_near``` requires that the minimum tolerance be ```DBL_MIN/DBL_EPSILON,``` which is approximately ```1.00208e-292```. That is, any two arguments whose difference is less than or equal to ```1.00208e-292``` are considered to be within tolerance, regardless of the specified tolerance.
+
+Before thinking that doubles should be compared to a tolerance smaller than ```1.00208e-292```, please consider that the ratio of the Planck length to the size of the observable universe is approximately ```1.8e-62```. Also consider that our minimum tolerance is ```5.4e-231``` times smaller than that. So, we think that'll probably be good enough in most cases.
```
ddj116, line 21: Screen Shot 2023-02-01 at 8 57 11 AM

trick_source/trick_utils/compareFloatingPoint/test/.gitignore
jmpenn: Fixed.

jdeans289 (approved):

PR #1453, opened 2023-02-07
SIM_aircraft Waypoint Update

Mrockwell2:
Updated the simulation to read in an input file of waypoint coordinates and images. When autopilot is toggled on, the simulation will follow a flight path set by these waypoints. The graphics client will use both the coordinates and images to display the waypoints on the map.


jmpenn:
I notice that there are a lot of png files in the **trick_sims/SIM_aircraft/images/** directory that are not related to the aircraft sim. Make sure this directory is tidied up so it doesn't inadvertently include files that aren't needed.

jmpenn:
In the README.md:
In the line:
```The desired heading is measured in radians, from -$\pi$ to $\pi$.```
Need to add a space between "-" and "$",  is so that it will render correctly.

Mrockwell2:
Both issues should be fixed.

jmpenn (approved):

PR #1456, opened 2023-02-13
1399 patch

### LINKED ISSUE #1399: Apparent race condition in checkpointing data recording groups
It seems like SIM_checkpoint_data_recording/RUN_test6 has revealed a race condition in checkpointing, which causes a segmentation fault when the checkpoint is reloaded. It appears that sometimes the variable name and/or variable alias in the data recording group does not get checkpointed correctly.

```
char* trick_anon_local_8[1];
char* trick_anon_local_9[1];
char trick_anon_local_10[23];
char trick_anon_local_11[1];
...
```
In a correct checkpoint:
```
trick_anon_local_8 = 
    {&trick_anon_local_10[0]};

trick_anon_local_9 = 
    {&trick_anon_local_11[0]};

```
In a buggy checkpoint:
```
trick_anon_local_8 = 
    {&trick_anon_local_10[0]};

trick_anon_local_9 = "";

```
For some reason, the variable_alias (sometimes variable_name as well, but less common) is not getting hooked up to the correct reference. This happens in about 1/10 runs, and only on RUN_test6

For now RUN_test6 is commented out. I've taken a few stabs at debugging this with no progress. The bug doesn't seem to happen when run with asan or tsan instrumentation, and I have not been able to identify where the issue is coming from. 

1. The issue is identified and fixed
2. SIM_checkpoint_data_recording/RUN_test6 passes consistently

jdeans289:
Closes #1399 

It appears that the checkpointing bug we were seeing in #1399 was due to the checkpoint logic reading uninitialized memory  that happens when an empty array of strings is creating and deciding to write it as a string instead of an array. This patch checks to make sure that when we write as a quoted string, we are in the correct dimension.

### DIFF
--- a/test_sims.yml
+++ b/test_sims.yml
@@ -240,32 +240,32 @@ SIM_checkpoint_data_recording:
       returns: 0
     RUN_test5/dump.py:
       returns: 0
-    # RUN_test6/dump.py:
-    #   returns: 0
+    RUN_test6/dump.py:
+      returns: 0
 
-    # RUN_test1/unit_test.py:
-    #   returns: 0
-    #   compare: 
-    #     - test/SIM_checkpoint_data_recording/RUN_test1/ref_log_foo.csv vs. test/SIM_checkpoint_data_recording/RUN_test1/log_foo.csv
-    # RUN_test2/unit_test.py:
-    #   returns: 0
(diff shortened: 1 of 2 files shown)


jmpenn (approved):
This is great! GOOD CATCH!

PR #1458, opened 2023-02-16
Check for unsupported types in data recording

### LINKED ISSUE #1440: Add a warning when trying to data record unsupported types

jdeans289:
closes #1440 

The previous resolution in #1441 failed to distinguish between arrays/pointers, which should not be data recorded and could stomp memory, and array elements, which should be data recorded. ~~Unfortunately, there does not seem to be a way to distinguish between an array and an array element using the REF2 data structure, so instead, if a variable with num_index != is found, use a regex to check if the name of the variable ends with an index [num]. I am not a fan of this solution, and I wouldn't be surprised if we decide to just roll back this check all together.~~

Found the solution in variable server code, should work properly now.

The purpose of this to distinguish between:
```
char * some_string = "A long string with many characters"; 
int some_arr[5];

// NOT allowed to be recorded:
some_string
some_arr

// is allowed to be recorded:
some_arr[0]
```


trick_source/sim_services/DataRecord/DataRecordGroup.cpp
```diff
@@ -339,6 +340,24 @@ int Trick::DataRecordGroup::add_change_variable( std::string in_name ) {
 
 }
 
+bool Trick::DataRecordGroup::isSupportedType(REF2 * ref2, std::string& message) {
+    if (ref2->attr->type == TRICK_STRING || ref2->attr->type == TRICK_STL || ref2->attr->type == TRICK_STRUCTURED) {
+        message = "Cannot Data Record variable " + std::string(ref2->reference) + " of unsupported type " + std::to_string(ref2->attr->type);
+        return false;
+    }
+
+    const static std::regex array_element_regex(".+(?:\\[\\d+\\])+$");
```
jdeans289, line 349: This will only match if the string ends with [num]. I am concerned that I could be missing some cases here that will cause unintended consequences. Another option is to just let this case go completely.

jdeans289: Found the actual way to do this from varserver code.

Issue #1460, opened 2023-02-21
VS_COPY_SCHEDULED fails unexpectedly when using custom job class order


cameronbracco:
When specifying a [custom job class order](https://nasa.github.io/trick/documentation/building_a_simulation/Simulation-Definition-File#specifying-scheduled-loop-job-class-order), the documentation states that all job classes "between automatic and automatic_last inclusive" are open for re-assignment.

If a user does not explicitly include the `automatic_last` job class in their custom job class order, variable server clients attempting to use  `VS_CODE_SCHEDULED` will fail to receive any data and an error will be printed in the simulation process:
```
Warning: Job trick_vs.vs.copy_data_scheduled of job class "automatic_last" is not handled by any scheduler
```

The default job class for this copy mode is set at https://github.com/nasa/trick/blob/master/share/trick/sim_objects/default_trick_sys.sm#L459.

Is there a reason to not use the `end_of_frame` job class to avoid this issue?

If the use of `automatic_last` intentional, I would recommend updating the documentation for [copy mode scheduled](https://nasa.github.io/trick/documentation/simulation_capabilities/Variable-Server#end-of-main-thread-execution-copy-mode--trickvscopyscheduled-or-1) to include a warning note, and if not, I would propose changing the default job class assigned.

jdeans289:
`VS_COPY_SCHEDULED` has to be an automatic job class because its cycle is not necessarily the same as the executive cycle. If you would like the copy to align with the frame, use `VS_COPY_TOP_OF_FRAME`.

This is definitely a bug, I'll look into it and hopefully patch it soon and follow up here. Until then, `VS_COPY_TOP_OF_FRAME` should be unaffected by this issue.

jdeans289:
This should now be resolved on master.

PR #1465, opened 2023-03-01
job_class_order no longer accidentally disables some system job classes

jdeans289:
Add automatic, automatic_last, and integ_loop job classes to scheduled loop even if left out of job_class_order spec. Includes docs and test.

When a user gives a specification for `job_class_order`, the order given will replace the entire scheduled job loop, and any job classes that are not included will not be handled by the scheduler. As a result, a user can accidentally disable the integ_loop and automatic jobs, which run events and the scheduled variable server mode.

A few important system job classes, like system_checkpoint and system_advance_time, are added back to the scheduled job loop even if they aren't specified by the user. This change adds automatic, automatic_last, and integ_loop jobs back to the scheduled loop in a similar way, and puts them in their default order. 

Addresses #1460


jdeans289:
After discussion with @jmpenn , this change also ensures that the user cannot reorder integ_loop, system_advance_sim_time, and other system job classes.

PR #1466, opened 2023-03-02
SIM_aircraft Waypoint Update

Mrockwell2:
Refactored the Waypoint class and implementation as well as added in basic file support in the graphics client.


trick_sims/SIM_aircraft/models/Aircraft/src/Waypoint.cpp
```diff
@@ -0,0 +1,89 @@
...
+    append(path);
+}
+
+void WaypointList::append(std::string path) {
+    std::ifstream in(path);
+
+    if(in.is_open()) {
+        std::string str;
+        while(getline(in,str)) {
+            int k = str.find(','), j = str.find(',',k+1);
```
jdeans289, line 65: I'd add a comment here that says the expected format of what this is parsing, something like
```
// Expected format: <north>,<west>,<img_file>
```

trick_sims/SIM_aircraft/models/Aircraft/include/Waypoint.hh
```diff
@@ -0,0 +1,36 @@
...
+#endif
+
+#include <string>
+#include <vector>
+#include <fstream>
+#include <iostream>
+
+class WaypointList {
+    public:
+    double north[ARRAY_SIZE];
```
jdeans289, line 18: Can we make this class a little more object-oriented? Make the arrays and length private members, and add getter and setter methods for them as needed, that way you could check length in this class instead of making the Aircraft class doing it. That way, if you later decide to do something like changing the arrays to use different data structure, you are able to do it by just changing this class and not changing Aircraft.

jdeans289 (commented):
Remove code related to the waypointInputFile option in AircraftDisplay.java around line 669.

This looks great!

My biggest concern having unique icons for each waypoint. It's a cool idea to be able to specify icons for each waypoint, but it looks like they are just copies of the same icon. It might be clunky for the user to have to choose an image file for each waypoint, especially if they have to be unique. This will be more of a concern when you implement adding waypoints. Could you implement a default icon as well as an option to specify a different image?

trick_sims/SIM_aircraft/models/Aircraft/include/Waypoint.hh
Mrockwell2: I went ahead and made the variables private.

trick_sims/SIM_aircraft/models/graphics/src/main/java/trick/aircraftDisplay/FileMenu.java
```diff
@@ -0,0 +1,5 @@
+package trick.aircraftDisplay;
```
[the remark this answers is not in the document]
Mrockwell2: I didn't even notice that this was here. Should be deleted now.

trick_sims/SIM_aircraft/models/Aircraft/src/Waypoint.cpp
Mrockwell2: This is a great idea. Comment added

PR #1468, opened 2023-03-03
Update Install-Guide.md

alexis-nicole:
Section 3 shows how to compile trick. However, it seems that the command for installing Trick is missing. I have added it. Please let me know if that is supposed to be there. Thanks


jdeans289:
Hi Alexis, 

Thank you for your contribution, but this is no longer the recommended way to install Trick. This omission from the documentation is intentional.

PR #1499, opened 2023-05-11
Add option to specify a C++ standard to ICG for parsing

jdeans289:
Add an option to ICG to control the C++ standard it parses against. Currently, the default is whatever the newest version supported by LLVM is. 

```
TRICK_CXXFLAGS += -std=c++11
```

Valid options are c++11, c++14, c++17, c++20. 

Also adds a convenience target to the root Makefile for the ICG executable.


alexlin0:
c++20 is a valid option for the newest compiler versions.

alexlin0:
I had a thought on this one.  Instead of creating a new argument, "icg-std", can we parse the command line arguments that look like "-std=c++??".    My assumption is that users specifying a c++ version for ICG would want that standard to be used for everything and would have the standard looking argument already in TRICK_CXXFLAGS.

jdeans289:
Yeah, that makes more sense. This issue was created specifically by request of a user who had to work with old code that would clog up the terminal with errors during ICG parsing if it used anything higher than C++11.

I don't remember what my initial reasoning was to create a separate flag, but I'll update it to just use the normal std flag

PR #1511, opened 2023-05-27
Revised Monte Carlo Tutorial Page

garshaam:
This fork adjusts the tutorial page to flow better with the prior pages. The idea was to be easily replicable without repeating prior information too much. It starts from the cannon_numeric simulation finished in "Dynamic Events - Making Contact"


jdeans289 (approved):
This looks great! The flow is much better than before.

Issue #1513, opened 2023-05-31
Variable Server byteswapping crashes the sim


alexlin0:
If you select the var_binary_nonames (or var_binary) and var_byteswap options on a variable server client, the code will byteswap too much data and will crash the sim.

jdeans289:
This is a known issue ( #1439 ), and there is a fix rolled in with other major refactors of variable server on the branch [variable-server-refactor-beta](https://github.com/nasa/trick/pull/1493). It hasn't been pulled into master yet, but will be eventually.

jdeans289:
Resolved with #1514

PR #1514, opened 2023-05-31
Variable Server byteswapping crashes the sim #1513

alexlin0:
The code to byteswap a variable server buffer has a comment saying there is a bug. The comment is correct.  The original code would swap the parameter and anything else that followed that parameter in the structure.  Crashes everywhere.  Created a new routine that byteswaps a single parameter. Strangely we didn't have such a routine until now.  Did some testing of doubles, floats, ints, shorts, and chars and all were swapped correctly.

### DIFF
--- a/include/trick/tc_proto.h
+++ b/include/trick/tc_proto.h
@@ -112,6 +112,7 @@ int tc_error(TCDevice * device, int on_off);
 int tc_dev_copy(TCDevice * dest, TCDevice * src);
 
 void *trick_bswap_buffer(void *out, void *in, ATTRIBUTES * attr, int tofrom) ;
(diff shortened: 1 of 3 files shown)


alexlin0:
I started looking at the variable server refactoring merge and it is very extensive.  We do need this in the short term and it only affects the one specific case in the variable server so it is a safer merge.  I would like to merge it first with the knowledge that it will be overwritten when the new variable server code comes online.

jdeans289:
Sounds good! I wrote a byteswap test that is currently disabled in this branch - would you mind applying this patch to enable it?

[enable_byteswap_test.patch.txt](https://github.com/nasa/trick/files/11615576/enable_byteswap_test.patch.txt)

PR #1518, opened 2023-06-07
Fixed an error with checkpointing change variables

Mrockwell2:


jmpenn:
Just add a test as we discussed.

PR #1527, opened 2023-06-28
Add issue templates

mcmuffin6o:
I tweaked the boilerplate for bug reports and feature requests provided by GitHub, and added a question template.

I'm hoping this will help users and developers alike in finding solutions more quickly.


sharmeye:
Thank you for your suggestion.  We may consider this route in the future, but for now we don't think adding this extra layer of formality is necessary for our project.

PR #1535, opened 2023-07-13
Additional debugging when adding monte carlo nodes

apshoemaker:
I am in the midst of setting up monte carlo simulations within a containerized kubernetes environment and had to work through a few different network-based gotchas.   It became helpful for me to fork the repo and add this simple additional debug value to the socket initialization error in order to track down a connection issue (which helped quite a bit!).  I thought it might be advantageous to add something _like_ this to the upstream code base, but most likely perhaps a much more though-through design (perhaps with some sort of explanation of error given the return code, etc).  

I mostly wanted to bring this to the attention of the community as a discussion point.  If there is interest and thought around what we would like to see here, I might be able to help contribute.

### DIFF
--- a/trick_source/sim_services/MonteCarlo/MonteCarlo_slave_init.cpp
+++ b/trick_source/sim_services/MonteCarlo/MonteCarlo_slave_init.cpp
@@ -34,10 +34,10 @@ int Trick::MonteCarlo::slave_init() {
 
     /** <li> Connect to the master and write the port over which we are listening for new runs. */
     connection_device.port = master_port;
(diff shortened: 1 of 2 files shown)


sharmeye:
Thanks for the input.  We'll run this through our testing, but the change is quite reasonable.  If you'd like to discuss more robust and thorough error reporting for Monte Carlo that you think might benefit users who are using a similar setup to yours, you can open some issues or contact us directly to discuss.

apshoemaker:
I am going to close this pull request and create a branched PR so that I can keep it clean from other debug code that you may not want to pull in.

Issue #1529, opened 2023-07-03
c++ instances allocated via MemoryManager do not call destructors when deleted


excaliburtb:
Similar to https://github.com/nasa/trick/issues/1301 but for a different reason. I found that if I declare a c++ class using memory manager (i.e. TMM_declare_var_s for my specific test), the alloc_info instance that gets allocated to track it always has the allocation set for TRICK_ALLOC_MALLOC. Since that enum `TRICK_ALLOC_MALLOC = 0`, it's either a mistake and is just resulting to the default.. OR, allocations are intentionally tracked as MALLOC instead of TRICK_ALLOC_NEW for some reason. git grepping through the code, I see no attempt to set the alloc_type to anything except for the DataRecordGroup. So, it doesn't matter which API I use, Trick will never treat it as anything other than a malloc.

sharmeye:
Thanks, we'll look into this and let you know.

jmpenn:
It looks like we might have just overlooked adding  "alloc_type = TRICK_ALLOC_NEW"
in the two occurrences where we set "language = Language_CPP" in :

void* Trick::MemoryManager::declare_var(TRICK_TYPE type,std::string user_type_name, int n_stars, std::string var_name, int n_cdims, int *cdims); 

,in MemoryManager_declare_var.cpp.

I'll add this, and test to ensure that it works as intended.

jmpenn:
Specifically, my proposed fix is:
In [MemoryManager_declare_var.cpp](https://github.com/nasa/trick/blob/master/trick_source/sim_services/MemoryManager/MemoryManager_declare_var.cpp)

After line 48, add:
TRICK_ALLOC_TYPE alloc_type;

After lines 128 & 141, add:
alloc_type = TRICK_ALLOC_NEW;
144: alloc_type = TRICK_ALLOC_NEW;

After line 158, add:
new_alloc->alloc_type = alloc_type;

excaliburtb:
that looks good to me

jmpenn:
Complication: It looks like the generated io_src allocation routine sometimes calls calloc, and sometimes it calls new. Need to do more digging.

excaliburtb:
hmm.. isn't there a io_src_*_destruct routine that gets generated? can't we use those?

PR #1539, opened 2023-07-25
Fix the alloc_type of allocations performed by declare_var. #1529

jmpenn:
declare_var() should indicate in ALLOC_INFO::alloc_type, how it has allocated the new object. From this we know how to delete it.

declare_var() can allocate an object in one of two ways: 1) using  malloc or calloc, or
2) using io_src_alloc_<CLASS>, which is generated by ICG.

PROBLEM 1 :
ALLOC_INFO::alloc_type was not properly set when allocating a composite type. It should have been set to TRICK_ALLOC_NEW to indicate that it was allocated by io_src_alloc_<CLASS>.

PROBLEM 2:
The enumeration value TRICK_ALLOC_NEW is mis-named and is misleading. The io_src_alloc_<CLASS> routines do not necessarily allocate memory using new. If the object is a POD type, it uses calloc.

SOLUTION:
So, THIS commit:

1) Renames TRICK_ALLOC_NEW to TRICK_ALLOC_IOSRC so as not be misleading.

2) In declare_var(), sets alloc_type = TRICK_ALLOC_IOSRC when allocating a composite type (TRICK_STRUCTURED).

### DIFF
--- a/include/trick/io_alloc.h
+++ b/include/trick/io_alloc.h
@@ -7,16 +7,38 @@
 #include "trick/attributes.h"
 #include "trick/parameter_types.h"
 
+/**
+ * enum TRICK_STCL - indicates whether or not the MemoryManager allocated the object.
+ * -------------------------------------------------------------------------------------------
+ * TRICK_LOCAL  - The memory manager allocated the object, via a form of declare_var..().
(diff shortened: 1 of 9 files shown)


jmpenn:
Closing this pull request without merging. I'll submit another merge request when I have a proper fix.

PR #1540, opened 2023-08-01
Add HOWTO for Python VIrtual Environment.

jmpenn:
Title says it all.


jmpenn:
Also updated test_linux.yml and code coverage.yml to fix a problem that arose on Oracle Linux 8.
Maven couldn't find the java compiler, because JAVA_HOME wasn't set.

Issue #1546, opened 2023-08-09
MonteVarCalculated variables that are of data type "int" cause incorrect data to be sent to child processes due to invalid memory access


nbrewerh:
In Trick::MonteVarCalculated::get_next_value() in MonteVarCalculated.cpp, there is a switch statement based on the monte carlo variable data type, but there is no code to handle TRICK_INTEGER or TRICK_UNSIGNED_INTEGER.

This causes a fallthrough to the TRICK_UNSIGNED_LONG case, where the pointer (which is intended to point to an int) is cast to a long pointer. This in turn causes invalid memory access when the pointer is dereferenced, and so incorrect data is sent to the child process.

jdeans289:
We're currently refactoring MonteCarlo, we'll keep an eye out on this. It seems like there should be a simple solution, we'll look into patching this. Feel free to make a contribution!

nbrewerh:
Created pull request
https://github.com/nasa/trick/pull/1547

jdeans289:
Closed by #1547

PR #1547, opened 2023-08-10
Fixed an issue where MonteVarCalculated variables of data type int were not passed correctly to child processes (#1546)

nbrewerh:


jdeans289:
Great catch, thank you for the contribution!
closes #1546

Issue #1559, opened 2023-08-29
Memory leaks in dllist.c


dandexter:
trick_source/trick_utils/trick_adt/src/dllist.c

Line: 270
Issue: In function DLL_InsertBefore*(), if the 'pos' and 'list' variables are null then the function returns before freeing 'newpos' that was just allocated resulting in a memory leak.
FROM:
```c++
DLLPOS newpos = (DLLPOS) malloc(sizeof(DLLNODE));
if (pos == NULL && list == NULL) {
    fprintf(stderr, "List and Position are NULL");
    return NULL;
}
``` 
TO:
```c++
if (pos == NULL && list == NULL) {
    fprintf(stderr, "List and Position are NULL");
    return NULL;
}
DLLPOS newpos = (DLLPOS) malloc(sizeof(DLLNODE));
``` 

----------
Line: 302
Issue: In function DLL_InsertAfter*(), if the 'pos' and 'list' variables are null then the function returns before freeing 'newpos' that was just allocated resulting in a memory leak.

FROM:
```c++
DLLPOS newpos = (DLLPOS) malloc(sizeof(DLLNODE));
if (pos == NULL && list == NULL) {
    fprintf(stderr, "List and Position are NULL");
    return NULL;
}
``` 
TO:
```c++
if (pos == NULL && list == NULL) {
    fprintf(stderr, "List and Position are NULL");
    return NULL;
}
DLLPOS newpos = (DLLPOS) malloc(sizeof(DLLNODE));
``` 

----------
Line: 379
Issue: In function DLL_AddHead*(), if the 'list' variable are null then the function returns before freeing 'newpos' that was just allocated resulting in a memory leak.
FROM:
```c++
DLLPOS newpos = (DLLPOS) malloc(sizeof(DLLNODE));
if (list == NULL) {
    fprintf(stderr, "List is NULL");
    return NULL;
}
``` 
TO:
```c++
if (list == NULL) {
    fprintf(stderr, "List is NULL");
    return NULL;
}
DLLPOS newpos = (DLLPOS) malloc(sizeof(DLLNODE));
``` 

----------
Line: 412
Issue: In function DLL_AddTail*(), if the 'list' variable are null then the function returns before freeing 'newpos' that was just allocated resulting in a memory leak.
FROM:
```c++
DLLPOS newpos = (DLLPOS) malloc(sizeof(DLLNODE));
if (list == NULL) {
    fprintf(stderr, "List is NULL");
    return NULL;
}
``` 
[...]

PR #1577, opened 2023-10-03
Fix a variety of bugs found in dllist.c while addressing Issue #1559.

jmpenn:
* Added More descriptive error messages.
* Wrote google-test based unit tests for the DLL List library.
* Deleted old, incomplete test program.
* Added new function DLL_ListContainsPos which determines whether the given list contains the node at the given pos.
* DLLFind bug: Added check to determine whether a compare function has been specified, and emit an error message if it hasn’t.
* DLL_FindIndex bug: Corrected bounds checking.
* DLL_GetAt bug: Added check to ensure that the specified pos is actually in the given list.
* DLL_SetAt bug: Added check to ensure that the specified pos is actually in the given list.
* DLL_RemoveAt: corrected logic mistake in NULL ptr check. Added check to ensure that the specified pos is actually in the given list.
* DLL_InsertBefore bug: element count not correctly updated.
* DLL_InsertAfter: next and prev ptrs not correctly updated, which corrupted the list.
* DLL_GetNext bug: logic error in NULL ptr checks. Added check to ensure that the specified pos is actually in the given list.
* DLL_GetPrev bug: logic error in NULL ptr checks. Added check to ensure that the specified pos is actually in the given list.
* DLL_AddHead: Fixed NULL check logic error.
* DLL_AddTail: Fixed NULL check logic error.


sharmeye:
I can think of at least one person who will be very happy with this update.  Everything looks good to me, and i like there's a unit test for this library now.

sharmeye (approved):
I don't see anything worrisome, looks good to me.

trick_source/trick_utils/trick_adt/src/dllist.c
```diff
@@ -19,12 +19,13 @@ DLLIST *DLL_Create(void)
 
 /* initializes a list */
 /* return: none */
```
hchen99, line 21: return: none

trick_source/trick_utils/trick_adt/src/dllist.c
```diff
@@ -34,15 +35,34 @@ void DLL_Delete(DLLIST * list)
     }
 }
 
+/* Determine whether the given list contains the node at the given pos.
+   If it does, return 1, otherwise return false.
```
hchen99, line 39: otherwise return 0 ;-)

trick_source/trick_utils/trick_adt/src/dllist.c
```diff
@@ -34,15 +35,34 @@ void DLL_Delete(DLLIST * list)
...
+        lpos = lpos->next;
+    }
+    if (lpos == NULL) {
+        return 0;
+    }
+    return 1;
+}
 
 /* returns number of elements in list */
 /* return: count */
```
hchen99, line 57: return: none

hchen99:
Really like those descriptive error messages

trick_source/trick_utils/trick_adt/src/dllist.c
jmpenn: Addressed this in commit below.

hchen99 (approved):

Issue #1616, opened 2023-11-10
Question : How to round trick view value


tastangh:
Hello everyone,
I am making a real-time simulation. During this simulation, I observe the data via trick_view. While observing the data, I want to see the data rounded to the nearest value, for example, I want the values ​​in the example below to appear on the view screen as 77.51, 7.44, 43.30, 29.03. While making this request, I want the variable type in the header or the data type used in my algorithm not to be changed, that is, it should be calculated as 77.50927363870733 in my algorithm and simulation, but it should appear as 77.51 on the view screen. Is there a way to do this easily?
![tastangh](https://github.com/nasa/trick/assets/54810017/1b43d384-e98a-4da8-9398-d2d3d22261a0)

best regards,

hchen99:
I don't believe so.

PR #1620, opened 2023-11-30
Format "Rounded" added to TVDouble for Issue #1616 Resolution

Fjolnirr:
In TV, TVDouble Format, `Rounded`  added. It can round up to `.2f`     
It is a workaround for  Issue #1616

### DIFF
--- a/trick_source/java/src/main/java/trick/tv/TVDouble.java
+++ b/trick_source/java/src/main/java/trick/tv/TVDouble.java
@@ -26,7 +26,14 @@ public class TVDouble extends VSDouble implements TrickViewFluent<TVDouble.Forma
                 return Double.toString(value);
             }
         },
-
+        
+        Rounded {
+            // Rounds the double value up to .2f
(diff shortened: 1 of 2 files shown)


sharmeye:
Thanks for the input.  The issue here is that rounding to two digits is not the only desired rounding accuracy.  We would consider something along these lines if the user could maybe set their desired rounding accuracy in the gui somewhere, dynamically updating the displayed values depending on what the user selected.

Fjolnirr:
Thanks for the comment. I was also aware that rounding to x digits would be more convenient but I wasnt sure where to add such gui element. I could also look into it to set the rounding digits somewhere in the gui.

alexlin0:
There used to be a way to specify the format of the numbers for the Y axis in the DP files.  It is still settable in the DP GUI, in the DP editor, when a plot is selected, a "Y axis format" field is settable.  It is written to the DP file, but I don't think it is parsed in the DPM code.

Issue #1626, opened 2023-12-19
Repurpose the "Sim/Real Time" field on the sim control panel


alexlin0:
The Sim/Real Time field is nearly useless.  It almost always reads 1.0.   It shows the desired real time clock ratio between sim and real time.  Nearly all users either run in straight realtime or non-realtime mode and the field reads 1.0.  It's only when someone uses the throttle does it change to something else.  I want to repurpose the field to show the actual sim/real time ratio.

PR #1627, opened 2023-12-19
Repurpose the "Sim/Real Time" field on the sim control panel #1626

alexlin0:
Added calculations in the realtime sync monitor job to calculate the average sim/realtime ratio for the past 100 frames.  This is saved to a new variable, actual_run_ratio.  Changed the sim control panel to display the actual_run_ratio value.

### DIFF
--- a/include/trick/RealtimeSync.hh
+++ b/include/trick/RealtimeSync.hh
@@ -104,6 +104,9 @@ namespace Trick {
             /** The clock time when the sim ended.  Used for total actual time calculation.\n */
             long long sim_end_time ;           /**< trick_units(--) */
 
+            /** The actual simulation time/wall clock time ratio\n */
+            double actual_run_ratio ;           /**< trick_units(--) */
+
             /**
              @brief This is the constructor of the RealtimeSync class.  It starts the RealtimeSync as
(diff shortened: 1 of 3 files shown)


hchen99:
Thanks, Alex! We're saving this for the first thing after new year.

hchen99:
Just to verify:

- This pull request repurposed Sim / Real Time field of Sim Control gui to show the average ratio saved as `actual_run_ratio` for past 100 frames.
- The user still uses the existing Throttle feature to change `rt_clock_ratio`. 
- The actual average ratio is calculated based on `rt_clock_ratio`.

What do you think if changing to only show `rt_clock_ratio` in Sim / Real Time Field especially for reflecting what is changed by user using Throttle when sim is not started?

alexlin0:
I believe that nobody uses rt_clock_ratio, sims are either run at 1.0x realtime or they are run non-realtime as fast as possible.   It is probably a feature that could be deleted and no one would miss it.

I assume we want to keep the feature.  In this case, I would not suggest using the field for two different values.  If we want to show the actual sim/real ratio and the rt_clock_ratio set by the user, then we should have 2 fields on the sim control panel.

PR #1628, opened 2023-12-19
Update Environment Variables to cover TRICK_SWIG_EXCLUDE

Mrockwell2:
I also shifted the TRICK_ICG_EXCLUDE section to be contiguous with the TRICK_EXCLUDE. This way it should be a bit easier to find the exclude tags.


hchen99:
Looks good. One minor input is to add wording for TRICK_EXCLUDE that it excludes both ICG and SWIG

hchen99:
Looks good to me. Thanks!

PR #1633, opened 2024-01-03
Created New Years Eve Ball Drop Simulation

ethanmaxey:
Happy New Years! A couple of months ago I was interviewed for a position to work on Trick, and I was tasked with creating my own simulation. I welcome any feedback.


trick_sims/SIM_balldrop/README.md
```diff
@@ -0,0 +1,45 @@
...
+## Running the Simulation
+
+1. Run the simulation .exe accoording the the input file.
+   
+```bash
+cd ../../../SIM_balldrop
+```
+
+```bash
+./S_main_*.exe RUN_test/input.py
```
markmaxey, line 38: I recommend eliminating the wildcard `*` in `./S_main_*.exe` by using the full file name.

trick_sims/SIM_balldrop/README.md
```diff
@@ -0,0 +1,45 @@
...
+```
+
+## Dynamics Model
+
+The force acting on our ball will be gravity only.
+
+To determine the position of the ball, we use the fundamental kinematics equations that $v = at\$ and $\text{pos} = \frac{1}{2}at^2 \$
```
markmaxey, line 45: Consider adding a description of each variable that plays into the physics of the simulation, the default for the variable, and a hint as to how to change the default if desired.

trick_sims/SIM_balldrop/models/graphics/src/main/java/trick/ballDisplay/BallDisplay.java
markmaxey, line 1: It appears you've followed the pattern of others by cloning a similar Java file and tweaking it to meet your needs.  That's fine.  Someone, at some point, will want to refactor the commonalities between all these Java files into reusable classes so that each new simulation doesn't need to reinvent the wheel by reusing common code.

sharmeye:
Thanks for the suggestion.  Hong and John noticed this pull request and brought it to my attention.  This sim appears to be similar to the existing cannonball sim, so I'm not sure that we need to have another gravity-driven spheroid simulation included in our sample sims area, but thank you for the contribution all the same.

PR #1636, opened 2024-01-09
Standardized error messaging in Sim Control Panel.

Mrockwell2:


hchen99:
Thanks Marcus! Thought this is for a larger scope initially. After looking further, realized the effort is for printing the error only to the status msg pane of the sim control panel when the connection to a variable server can't be established unless the sim control panel is in Lite mode, then use a popup window for the error msg. I am good with the idea. 

- Original:
   - Run `trick-simcontrol` without specifying host and port, the error msg goes to a popup window regardless of the GUI mode (either Full or Lite)
   - Run `trick-simcontrol host port` with specified host and port, the error msg goes to screen if the host is not right or the port is not right for connection.

- New:
   - Run `trick-simcontrol` without specifying host and port
      - Lite: the error msg goes to a popup window 
      - Full: the error msg goes to the status msg pane
   - Run `trick-simcontrol host port` with specified host and port, the error msg is intended to go to the status msg pane if the host is not right or the port is not right for connection.
      - Marcus, might be just me, would you please check this as I am getting NullPointerException because the status msg pane is not ready yet for displaying error msg when the sim control panel is launched this way? Thanks!

hchen99:
All looks good to me.

Issue #1637, opened 2024-01-12
S_job_execution enhancement 


amishscientist:
It would be nice if there was more thread info in the thread section of the S_job_execution file.  

For example it would be nice to know the rates for the child threads would help a lot into figuring out quickly if sims have child type or rate mismatches. 

This is what it is showing me now

Trick::Threads (Child_2)
    process_type = asynchronous
    trigger type = mutex
    number of scheduled jobs = 22
    from Trick::ThreadBase
    pthread_id = 140326706603776
    process_id = 1417105
    rt_priority = 0
    cpus = none assigned
Trick::Threads (Child_3)
    process_type = scheduled
    trigger type = mutex
    number of scheduled jobs = 40
    from Trick::ThreadBase
    pthread_id = 140326698211072
    process_id = 1417107
    rt_priority = 0
    cpus = none assigned

hchen99:
Currently the cycle time is printed to S_job_execution only for PROCESS_TYPE_AMF_CHILD:
-  If you have a PROCESS_TYPE_AMF_CHILD thread and its cycle time is set to 0.1, you should see following in S_job_execution:
   `process_type = asynchronous must finish with amf_cycle = 0.1`

hchen99:
Updated to dump cycle time for all types. However, if there is a particular reason for only dumping cycle time for PROCESS_TYPE_AMF_CHILD, the pull request would get rejected.

PR #1638, opened 2024-01-15
S_job_execution enhancement #1637 

hchen99:
Dump cycle time for all types instead of just PROCESS_TYPE_AMF_CHILD.


jmpenn (approved):
Very straight forward.

PR #1639, opened 2024-01-23
Fix the alloc_type of allocations performed by declare_var. #1529

jmpenn:
This is a re-do of PR 1539 associated with 1529.


hchen99:
MM declare var sets alloc_type and the destructor gets called when delete_var, associated unit shows the intention. All looks good.

jmpenn:
Also see Issue #1529 for additional comments.

PR #1642, opened 2024-02-07
TV VariableTable precision column added to change significant figures after dot(.)

yusufcananar:
TV **Variable Table** precision column added to change significant figures after dot(.).
It only applies for `TVDouble` and `TVFloat` type of variables, any other format only shows "--". 
Every variable precision is "--" by default. Until selection, variable shows all figures it has.
This feature is related to #1620 and #1616 .
The feature is a generic solution to change fractional part precision  [1-15]. 
I find practical to use a comboBox for precision since there can only be limited number of significance.

trick_tv_precision


sharmeye:
I think we'd like this if it were an option that you could select in the view dropdown menu, and it would be unselected by default.  Then if the user selected this "Max Precision" option in View, the extra precision column would be added to the panel as you've shown above.

Issue #1656, opened 2024-02-24
Deadlock occurs between variable server creating threads and main thread shutdown


cameronbracco:
## Overview
We recently upgraded to Trick version 19.7.1 from version 19.4.0 and we began running into a regular failure where the trick sim would fail to shut down properly and hang infinitely. We root caused this to a deadlock between the main thread and the `VariableServerListenThread` during shutdown. We believe this was introduced as part of the new thread management/shutdown behavior in https://github.com/nasa/trick/pull/1448.

## Investigation
After running our trick sim with `gdb`, we discovered that the main thread was stuck in the shutdown loop while it was attempting to join all the `SysThread`s. More specifically, the main thread was stuck joining the `VariableServerListenThread` which is in turn handling a new connection request. At this point in time the main thread [holds the lock](https://github.com/nasa/trick/blob/19.7.1/trick_source/sim_services/ThreadBase/SysThread.cpp#L56) for the `list_mutex()`, which controls access to `all_sys_threads()`. When the `VariableServerListenThread` creates a `VariableServerSessionThread` it attempts to [add that object into `all_sys_threads()`](https://github.com/nasa/trick/blob/19.7.1/trick_source/sim_services/ThreadBase/SysThread.cpp#L42) , which means it must acquire `list_mutex()` . This creates a deadlock where the main thread holds the `list_mutex()` and is waiting to join the `VariableServerListenThread`, which is in turn blocked on acquiring to lock for `list_mutex()`.

We reached this conclusion by following stack traces. When the process deadlocks and we hit `Ctrl + C` in `gdb`, we can see the following stack trace:
```
Thread 1 "S_main_Linux_8." received signal SIGINT, Interrupt.
0x00007f0f085ce6cd in __pthread_timedjoin_ex () from target:/lib64/libpthread.so.0
(gdb) where
#0  0x00007f0f085ce6cd in __pthread_timedjoin_ex () from target:/lib64/libpthread.so.0
#1  0x0000000003c93447 in Trick::ThreadBase::join_thread (this=0x7a742b0 <trick_vs+176>) at ThreadBase.cpp:320
#2  0x0000000003c916fb in Trick::SysThread::ensureAllShutdown () at SysThread.cpp:65
#3  0x0000000003c34152 in Trick::Executive::shutdown (this=0x7a71f28 <trick_sys+168>) at Executive_shutdown.cpp:81
#4  0x0000000003d28695 in master (nargs=4, args=0x7fffea970f68) at master.cpp:47
[...]

PR #1657, opened 2024-02-25
Deadlock Fix

drgardner43:
This is one (of probably many) potential solutions to https://github.com/nasa/trick/issues/1656 by doing the following:

* Switching the list mutex to be a PTHREAD_MUTEX_NORMAL type, which allows us to implement mutex timeouts
* Implementing some basic mutex timeouts around accessing the list mutex
* Adding a new boolean flag `shutdown_initiated` to signal that we are in the process of shutting down and joining threads

This branch also currently has my test setup I used to replicate this in my at home box (WSL Ubuntu). As this is one of probably many different solutions here for this issue, i'll clean up branch if repo maintainers want to move forward and revert the test setup changes. If not - the test setup is here for replication :)

cheers


sharmeye:
Thanks for the suggestion.  We're working on a fix that will be implemented slightly differently.  It will be released with the next minor Trick release, so please let us know if the issue persists after the next release.

PR #1662, opened 2024-03-06
TrickOps: Use usedforsecurity=False when calling hashlib.md5 for compare mechanism

### LINKED ISSUE #1661: Trickops: Use usedforsecurity=False when calling hashlib.md5 for compare mechanism
Discovered as part of our upgrade of CML (Common Model LIbrary) to a Rocky 8 machine with FIPS mechanisms enabled, the `TrickWorkflow.py` compare mechanism will throw exceptions that look like this:
```
  File "./hashlibtest.py", line 5, in <module>
    h = hashlib.md5(open(__file__).read()).hexdigest()
ValueError: [digital envelope routines: EVP_DigestInit_ex] disabled for FIPS
```

Simple fix is to explicitly denote `usedforsecurity=False` in the call. This hash mechanism is only used to compare baseline data to test data within the TrickOps architecture.

ddj116:
This clears the "ValueError: ...  disabled for FIPS" exception for systems with FIPS mechanisms enabled. This is safe because hashlib is only used to compare baseline data to test data in trickops, it's not used for any security mechanisms.

FYI @hchen99 @jmpenn @sharmeye 

Closes #1661

### DIFF
--- a/share/trick/trickops/TrickWorkflow.py
+++ b/share/trick/trickops/TrickWorkflow.py
@@ -1284,8 +1284,8 @@ class TrickWorkflow(WorkflowCommon):
                   self.status = Job.Status.FAILED
             if self.missing:
                 return  self.status
-            if (hashlib.md5(open(self.test_data,'rb').read()).hexdigest() !=
-               hashlib.md5(open(self.baseline_data,'rb').read()).hexdigest()):
+            if (hashlib.md5(open(self.test_data,'rb').read(), usedforsecurity=False).hexdigest() !=
+               hashlib.md5(open(self.baseline_data,'rb').read(), usedforsecurity=False).hexdigest()):
                 self.status = Job.Status.FAILED
             else:
                 self.status = Job.Status.SUCCESS


hchen99:
Thanks for the PR. Did some research and the solution seems to be the right way to fix it. 

According to https://stackoverflow.com/questions/77425682/what-is-the-point-of-usedforsecurity:

One aspect of FIPS is restricting the hash functions you are allowed to use. In particular, MD5 is not allowed under FIPS. When you use MD5 in a FIPS environment, you will encounter an error. That is what the introduction of `usedforsecurity` is supposed to fix: give you an escape hatch in the case that you truly want to use MD5 in a FIPS environment. The parameter is designed to be specified at each call site so it can be audited on a case by case basis.

There seems to be confusion [on many sides](https://github.com/python/cpython/issues/53462#issuecomment-1093510157) that `usedforsecurity` has anything to do with security. It's not. Having it set to False doesn't reduce your security. On FIPS enabled environments, it does however switch your implementation of the hash function to one that was explicitly certified.

In conclusion, for all intents and purposes, the parameter might as well have been called `exceptionforfips` because that's the singular purpose it serves: as an escape hatch if you happen to work under a FIPS environment and still need to use a FIPS non-compliant hash.

ddj116:
@hchen99 Any objection to me tacking on a quick fix for #1663 to this PR before we merge? I can do a separate PR for that if you prefer.

hchen99:
> @hchen99 Any objection to me tacking on a quick fix for #1663 to this PR before we merge? I can do a separate PR for that if you prefer.

If you can do a separate pull request, that'd be great. Thanks!

PR #1664, opened 2024-03-07
Protect TrickOps progres bar from divide by zero

### LINKED ISSUE #1663: TrickOps: Protect for divide by zero in progress bar computation
First noticed in testing the unit sims that ship with the CML project, the traceback when it occurs looks like this:

```
  File "/deps/trick/share/trick/trickops/WorkflowCommon.py", line 801, in execute_jobs
    curses.wrapper(execute)
  File "/usr/lib64/python3.6/curses/__init__.py", line 94, in wrapper
    return func(stdscr, *args, **kwds)
  File "/deps/trick/share/trick/trickops/WorkflowCommon.py", line 745, in execute
    job.get_status_string() + '\n\n', color)
  File "/deps/trick/share/trick/trickops/WorkflowCommon.py", line 329, in get_status_string
    return self._running_string()
  File "/deps/trick/share/trick/trickops/TrickWorkflow.py", line 1499, in _running_string
    self._connected_bar())
  File "/deps/trick/share/trick/trickops/TrickWorkflow.py", line 1543, in _connected_bar
    progress = self._tics.value / self._terminate_time.value
ZeroDivisionError: float division by zero
```
Unclear under what conditions this can happen, probably when a small unit sim is able to connect to the var server before time advances, which I'm guessing is rare.  Even without understanding how it can occur, the resolution is probably to check for divide by zero in the computation of `progress` so it's not possible.

Trick Team - I can tackle this issue when I've got the bandwidth.

ddj116:
Closes #1663

### DIFF
--- a/share/trick/trickops/TrickWorkflow.py
+++ b/share/trick/trickops/TrickWorkflow.py
@@ -1540,7 +1540,10 @@ class SingleRun(Job):
           self._sim_time(), self._average_speed())
 
     def _connected_bar(self):
-        progress = self._tics.value / self._terminate_time.value
+        if self._terminate_time.value <= 0.0:
+          progress =  0.0
+        else:
+          progress = self._tics.value / self._terminate_time.value
         return create_progress_bar(
           progress, '{0:.1f}%'.format(100 * progress))
 


ddj116:
@hchen99 per your request here's the MR with only this change. Since TrickOps is now a critical part of our workflow, both this and #1662 will be needed for EG to move forward with Trick on Rocky 8 Linux, which is coming to our lab likely in early April.

hchen99:
Thanks for the PR. It looks very straightforward. Will try to merge both of your PRs if not this week probably early next week.

Issue #1666, opened 2024-03-12
ACTION REQUIRED: Critical Vulnerability Detected Please Fix ASAP


rubioJeffery:
Please notify NASA ARC nasa-data@lists.arc.nasa.gov once this is remedied.

PR #1667, opened 2024-03-13
ACTION REQUIRED: Critical Vulnerability Detected Please Fix ASAP #1666

hchen99:


irowebbn:
How can we know that jar that was committed actually has the JUnit patch?

sharmeye:
@irowebbn, thanks for the question.  If this is about the JUnit vulnerability CVE-2020-15250, that's a non-issue here because the bsaf library isn't being used in Trick in any way that the vulnerability could be exploited.  Also, our pom file uses the patched version for JUnit already.  I think we're good here, but thanks for the sanity check.

PR #1670, opened 2024-03-18
TV VariableTable max precision column option

yusufcananar:
View Dropdown menu has Max Precision option which is not selected by default.
If user selects this option Variable Table will show the precision column that has the following features:

- TV Variable Table precision column added to change significant figures after dot(.).
- It only applies for TVDouble and TVFloat type of variables, any other format only shows "--".
- Every variable precision is "--" by default. Until selection, variable shows all figures it has.
- The feature is a generic solution to change fractional part precision [1-15].

This feature is related to #1642  
I find practical to use a comboBox for precision since there can only be limited number of significance.

Example images of the feature;
trickTVMaxPrec
tvMaxPrecClosed


Fjolnirr:
Is there any progress on this PR? @sharmeye

dbankieris:
Although I don't actively develop Trick anymore, I stumbled across this PR today, and, as the original Trick View author, was compelled to offer some feedback:
1. Column visibility is already controlled via a widget in the top-right corner of the table. I do not recommend adding another method, despite Sean's [comment](https://github.com/nasa/trick/pull/1642#issuecomment-1986020367). You could still start with the precision column hidden, I think.
2. `precision` is meaningless for all but `TVFloat` and `TVDouble`, yet it has been added to every TV type. A better design would avoid what is essentially dead code.
3. Setting precision might be a better feature for the Trick Variable Server rather than Trick View. Then, _all_ Variable Server clients could benefit from it.

image

sharmeye:
I realize we never actually circled back to this (now a year later).  We've decided to add a float- and double-precision command to the Variable Server to set the output precision for the VS client.  This is less invasive and more flexible.

yusufcananar:
Did you make any progress about the variable server output precision feature? If not, I can add the feature to the variable server. Also, will there be an option to send this command to the variable server from the TV UI? @sharmeye

PR #1674, opened 2024-03-20
Add trickeifed classes.resource to sie generation after cp.

hchen99:


excaliburtb:
confirmed this fixes the issue for trickified JEOD

hchen99:
Glad to hear. Thanks for confirming!

libexec/trick/sie_concat
```diff
@@ -14,6 +21,17 @@ while(my $line = <$classes_resource>) {
 
 close($classes_resource);
 
+# Add trickified classes.resource if available
+foreach my $path ( @trick_python_paths ) {
+	my $trickified_dir = dirname($path);
+	open(my $classes_resource, "<", "$trickified_dir/build/classes.resource")
```
excaliburtb, line 27: I just ran this for a different sim and this `die` command will exit this script for any directory listed in TRICK_PYTHON_PATH that doesn't have a build/classes.resource. We actually want to keep trying the next directoy.

PR #1675, opened 2024-03-21
python3 GIL locking solution

excaliburtb:
Implemented use of the python GIL API to replace the less portable mutex solution.


excaliburtb:
i titled it python3 but it still works with 2.7.18 as well

hchen99:
Thanks for the fix. According to https://docs.python.org/3/c-api/init.html#non-python-created-threads, for non python created threads, using GIL is the way suggested.

PR #1679, opened 2024-03-27
write template directive before template instance

excaliburtb:
in newer swig versions. 4.2.1 confirmed.. maybe issue with any 4.x version, a template with a private copy constructor or operator will generate setters for the instance because swig doesn't know what the template is. swig documentation says to write the `%template` directive first. 

This PR does that though I'm no expert on regex. feel free to rework as needed but this PR gets the job done for swig 4.2.1 and is backwards compatible with 3.0.12

See attachment to reproduce the error:

[template_error.zip](https://github.com/nasa/trick/files/14774719/template_error.zip)


sharmeye:
@alexlin0 you wrote this in perl, this is your fault.  Please read this and tell us what it says

PR #1682, opened 2024-03-28
Trick prints warning for checkpoints outside of freeze mode.

Mrockwell2:


excaliburtb:
i don't think this is true for generic cases. if you have some fully async threads running, then it might be a problem. otherwise, it is well-defined behavior. Perhaps you have a specific use case or example that illustrates the issue?

Mrockwell2:
> i don't think this is true for generic cases. if you have some fully async threads running, then it might be a problem. otherwise, it is well-defined behavior. Perhaps you have a specific use case or example that illustrates the issue?

When you checkpoint a running sim without freezing and reload that checkpoint, the data recording log resumes on the next frame. So far that's the most glaring thing I've found, but I'll keep looking.

Mrockwell2:
The issue fix was better implemented in another branch. See this pull request (#1727 )

PR #1694, opened 2024-04-16
Speed up Trick::ScheduledJobQueue::push

### LINKED ISSUE #1693: Trick::ScheduledJobQueue::push has O(n) complexity
`Trick::ScheduledJobQueue::push` scans linearly through a sorted list and copies items one by one.

This takes O(n) time and can be a significant bottleneck when there are many jobs.

The scan can be replaced with [`std::upper_bound`](https://en.cppreference.com/w/cpp/algorithm/upper_bound) and a suitable comparison function, making the complexity O(log n).

It also always allocates a new array of `JobData*`, so the entire array is always copied. This could be avoided by using `realloc` and `memmove`, resulting in fewer copies of larger chunks of memory.

EDIT: `Trick::ScheduledJobQueue::remove` has the same issue, but it's less impactful because removing jobs is less common than creating them.

EDIT 2: It actually has to be `std::upper_bound` instead of `std::lower_bound` in order to preserve insertion order.

iamthad:
Fixes #1693

### DIFF
--- a/trick_source/sim_services/ScheduledJobQueue/ScheduledJobQueue.cpp
+++ b/trick_source/sim_services/ScheduledJobQueue/ScheduledJobQueue.cpp
@@ -1,8 +1,9 @@
-
+#include <algorithm>
 #include <iostream>
 #include <sstream>
 #include <stdlib.h>
 #include <stdio.h>
+#include <string.h>
 
 #include "trick/ScheduledJobQueue.hh"
 #include "trick/ScheduledJobQueueInstrument.hh"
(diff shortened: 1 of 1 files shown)


trick_source/sim_services/ScheduledJobQueue/ScheduledJobQueue.cpp
```diff
@@ -46,72 +71,24 @@ Trick::ScheduledJobQueue::~ScheduledJobQueue( ) {
...
-            }
-        } else if ( list[ii]->job_class < new_job->job_class ) {
-            new_list[jj++] = list[ii] ;
-        } else {
-            new_list[jj++] = new_job ;
-            break ;
-        }
+    JobData ** new_list = (JobData **)realloc(list, (list_size + 1) * sizeof(JobData *)) ;
+    if (!new_list) {
+        abort();
```
iamthad, line 77: This could be replaced with some other logic, but the old code wasn't handling allocation failure at all.

sharmeye:
We checked this out and are satisfied it's working as expected.  Thank you for the contribution, we appreciate the help.

PR #1696, opened 2024-04-18
Fix bug dereferencing namespaces in convert_swig

excaliburtb:
Changed to many perl arrays to lists.


excaliburtb:
[template_error_v3.zip](https://github.com/nasa/trick/files/15031496/template_error_v3.zip)

Added test that handles both the private swig4.2 case and the template inheritance issue the previous convert_swig change introduced.

Issue #1700, opened 2024-04-26
Support RHEL8 on PowerPC architecture


alexlin0:
Found this in a round-about way.  I was trying to compile Trick on a PowerPC machine running Redhat 8.  Could not compile Trick.  I found that sections of code with "#if __linux" were not getting included.  I found that __linux__ was used on my machine instead.  But while experimenting I found that __linux is defined if we don't use the --std=c++11 argument.  It's been years since this was added.  The default version of g++ 8 (the default version that comes with RHEL8) is now c++14.  Removing this argument throughout Trick makefiles will compile Trick code to a more recent c++ standard and fixes my build.

alexlin0:
Nope, can't do it.  Found a couple of lines on some machines that still require the use of --std=c++11.  Changing the title to my underlying purpose, support RHEL 8 on PowerPC.

Accomplish this by changing __linux to __linux__ in preprocessor statements.

dandexter:
I know you are trying to get Trick to compile on RHEL 8 on PowerPC, but what about emulations that need to be compiled with -std=c++11? Would there be any problem with the Trick executive and supporting trick code that gets added to the simulation when the sim is compiled with -std=c++11?  If you find a way that support c++11 that does not depend on __linux then perhaps this is not even a problem.

alexlin0:
I thought I read that the default compiler versions on RHEL and Mac were already supporting c++ 14 or newer by default.  I thought that removing c++11 would work without issues.  I was wrong.  Somewhere there is still a dependency.  I reverted that change and am trying a different way.

dandexter:
The issue I'm thinking of is IEEE 1516-2010 uses deprecated APIs and that causes compile problems for c++14 and I have to set -std=c++11 for RHEL 8 or newer. I just wanted to share that some users may have to use c++ 11 to compile simulations and I wanted to make sure there would not be any issues for any fix you come up with.

dandexter:
The IEEE 1516-2010 provides APIs and libraries that are targeted to the c++ 11 standard. The next version of the HLA standard (i.e. HLA 4) is going through balloting right now but it is my understanding that the APIs and libraries will also be c++ 11. So no we can't change the deprecated APIs and it will be that way for the foreseeable future.

dandexter:
I think it will be okay tfor Trick to be compiled without specifying a C++ standard, but when we compile the simulation we do.

PR #1702, opened 2024-04-26
Support RHEL8 on PowerPC architecture #1700

alexlin0:
Did a global replace of __linux with __linux __


sharmeye:
Just curious, why did you replace instead of making it optional via "or"?  #if( __linux | __linux__) for example?

alexlin0:
I wanted to minimize the number of variables we are looking for.

Issue #1461, opened 2023-02-22
Only allow dumping and loading checkpoints in freeze mode


jdeans289:
If you dump or reload a checkpoint while not in freeze mode, you can have unexpected results. Currently, the sim control gui enforces this by greying out buttons, but there is no mechanism preventing a checkpoint dump or reload in run mode in Trick itself, so manually calling dump_checkpoint in run mode can result in confusing bugs. 

This issue can be closed when:
- Trick prints a warning and refuses to dump/load a checkpoint if not in freeze mode
- This behavior is tested and documented

PR #1727, opened 2024-05-31
Prevent Dumping and Loading Checkpoints During Run Mode

Mrockwell2:
Solves issue #1461 while also allowing for scheduling checkpoints at future times and using trick.add_read to load checkpoints at specific times.


Mrockwell2:
Just found that this might break Master-Slave. The Slave can mimic the Master automatically when it dumps or loads a checkpoint. Going to have to look at how the Master dumps and loads checkpoints. I'm also going to create an issue that MasterSlave has no tests.

sharmeye:
Closing and moving to internal branch to update and stress-test

PR #1740, opened 2024-07-11
Restore ROBoT image in Related Projects

jefferyjohn:
Updates image reference to ROBoT.

Current: 

image

Proposed:

image


sharmeye:
Thanks for pointing this out.  We are actually planning to redo this entire section, since a lot of the projects that Trick is currently used on are not represented, and some of the project listed in this section are no longer active.  When we refactor this section we'll make sure all the links actually work.  Thanks again.

PR #1752, opened 2024-07-30
Correct convert_swig regex for qualified templates

### LINKED ISSUE #1751: convert_swig's regex for matching qualified template instantiations is incomplete
[This line](https://github.com/nasa/trick/blob/92b0168b7bda0fbcc4ab026e0f1efa4d05fb6b7d/libexec/trick/convert_swig#L718) is looking for template instantiations that are qualified with the scope resolution operator (`::`). The pattern matches instantiations with only a single operator (such as `bar::Foo`) but not those with multiple operators (such as `baz::bar::Foo`). I believe the correct pattern is `^\w*(::\w+)+<`

dbankieris:
Remove dead code

Closes #1751

### DIFF
--- a/libexec/trick/convert_swig
+++ b/libexec/trick/convert_swig
@@ -57,7 +57,6 @@ my %sim ;
 my %out_of_date ;
 my ($version, $thread, $year) ;
 my %processed_templates ;
-my $global_template_typedefs ;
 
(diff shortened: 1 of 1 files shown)


dbankieris:
Yeah, you can just close this one.

PR #1755, opened 2024-08-05
Wiki button

Pherring04:
Added a button to the sim control panel that will link to trick documentation


sharmeye:
One less reason people can claim they don't know where the documentation is.

PR #1757, opened 2024-08-07
Updated to use default-jdk package instead of openjdk-11-jdk for ubuntu.

hchen99:
Using `default-jdk` instead of a specific JDK version so that if there is JRE installed by default, it'll be consistent with the JDK. 
Ubuntu 24 has JRE (version 21) headless set up by default in `/usr/bin`.
Installing openjdk-11-jdk in ubuntu24 does set up openjdk-11-jre however in `/usr/lib/jvm/java-11-openjdk-amd64/bin` instead of `/usr/bin`. openjdk-11-jre doesn't replace the JRE headless in `/usr/bin` probably because the default one is a newer version. 

If using openjdk-11-jdk package in ubuntu24, the user needs to set their environment path to use openjdk-11-jre (`java` command for running Java GUIs) first otherwise, they'd get an error by running Java GUIs using headless JRE from `/urs/bin` by default.
Using `default-jdk` works for ubuntu 18, 20, 22, and 24. Using openjdk-11-jdk works for 18, 20, 22 without extra work as their default `java` version is 11 and openjdk-11-jre can be set to the default path.

Trick CI uses ubuntu22 with `default-jdk` package btw.

### DIFF
--- a/docs/documentation/install_guide/Install-Guide.md
+++ b/docs/documentation/install_guide/Install-Guide.md
@@ -198,6 +198,8 @@ libgtest-dev default-jdk zip
 export PYTHON_VERSION=3
 ```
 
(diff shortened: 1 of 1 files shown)


sharmeye:
@jmpenn demands **absolute clarity**, and that includes his particular version of grammar to be dogmatically adhered to.

jmpenn:
@sharmeye I don't claim good grammar to be **my** version. I know it's hard, but it'll be OK.

PR #1760, opened 2024-08-13
Need a bit more info in Integrator class

excaliburtb:
Added "target_integ_time" to Integrator class and C APIs. It can be useful to know the final integration time even during intermediate steps.


sharmeye:
This is pretty low impact, so I don't really see a problem with this.  Your most polite PR ever probably, and also it _appears_ to be the least troublesome.

PR #1762, opened 2024-08-15
added SIM_singlerigidbody

bnwalla1:


sharmeye:
There are a few sections of the README that include information that's already contained in the tutorial.  We would like that removed before merging.  Specifically the sections **Building the Sim**, **Running the Sim**, and **Recorded Data** contain info that is already covered elsewhere.

sharmeye:
It appears the "sims" directory located in your SIM_singlerrigidbody directory is redundant and contains a lot of autogenerated files.  Can you please delete it?

trick_sims/SIM_singlerigidbody/models/src/body.cpp
```diff
@@ -0,0 +1,353 @@
...
+			massmatrix[i][j] = mass; 
+
+	for(int i = 1; i<2; i++)
+		for(int j = 1; j<2; j++)
+			massmatrix[i][j] = mass; 
+
+	for(int i = 2; i<3; i++)
+		for(int j = 2; j<3; j++)
+			massmatrix[i][j] = mass; 
+
```
sharmeye, line 147: Same comment as above about unnecessary nested for-loops

trick_sims/SIM_singlerigidbody/models/src/body.cpp
```diff
@@ -0,0 +1,353 @@
...
+	for(int i = 3; i<6; i++)
+		for(int j = 0; j<3; j++)
+			mat_mass[i][j] = m_CM_skew[i-3][j];
+
+	for(int i = 3; i<6; i++)
+		for(int j = 3; j<6; j++)
+			mat_mass[i][j] = inertia_matrix[i-3][j-3];
+
+};
```
sharmeye, line 166: If all of these values are constant, is there a reason that this routine is called in the derivative loop instead of at init?  There's no reason, if the mass matrix is constant, to continually reassign and recalculate the matrix.  It can just be done at init.

trick_sims/SIM_singlerigidbody/models/src/body.cpp
```diff
@@ -0,0 +1,353 @@
...
+	
+	// Solving a = b * M^-1
+	
+	// Temporary vector for Choleski 
+	double tmp_vec[6] = {0.0, 0.0, 0.0, 0.0, 0.0, 0.0};
+
+	// Dynamic memory set up
+	double **mat_mass_dyn;
+	double **mat_L;
+	mat_L = new double*[6];
```
sharmeye, line 271: Why are you repeatedly allocating and de-allocating the same memory during the derivative jobs?  These matrices are always the same size, and could be define in the header and allocated once during init and de-allocated at shutdown.

trick_sims/SIM_singlerigidbody/models/src/body.cpp
```diff
@@ -0,0 +1,353 @@
...
+
+};
+
+/***************************************************************************
+FUNCTION: BODY::derivative()
+PURPOSE: (Calls all functions in desired order for calculations)
+***************************************************************************/
+void BODY::derivative() {
+
+	mass_matrix();
```
sharmeye, line 303: see comment above about moving mass matrix calculation out of derivative job call

trick_sims/SIM_singlerigidbody/models/src/body.cpp
```diff
@@ -0,0 +1,353 @@
...
+	eq_solver();
+
+};
+
+/***************************************************************************
+FUNCTION: BODY::integrate()
+PURPOSE: (Sets up trick integration)
+***************************************************************************/
+
+#ifndef IN_MAKE
```
sharmeye, line 319: What is this #define for?

Issue #1777, opened 2024-09-09
Fix issue with keeping track of already deleted memory.


excaliburtb:
Started to see:

`MemoryManager:WARNING:The MemoryManager cannot delete memory at address [0x600002e2c600] because it has no record of it. Furthermore, the MemoryManager has not recently deleted it while resetting memory for a checkpoint reload.`

mostly on Mac. Some destructors call delete_var and Trick has to keep track of them during reset_memory(). The issue is this is only partially taken care of.

PR #1778, opened 2024-09-09
Add deletions to tracking list if malloc or new

excaliburtb:
fixes https://github.com/nasa/trick/issues/1777


sharmeye:
We're gonna merge this, but if you burn us again with one of these we're no longer accepting PRs from you without substantial verification that it isn't some hidden virus.

excaliburtb:
> We're gonna merge this, but if you burn us again with one of these we're no longer accepting PRs from you without substantial verification that it isn't some hidden virus.

did you have to write a unit test yourself? no you did not.

PR #1790, opened 2024-09-28
Added passing arguments to constrcutors when allocating with Trick Memory Manager.

M-Herr:
## Goals

1. Let users call constructors that take arguments when allocating with the trick memory manager.
2. Make this available in both C++ and Python

## New Things

1. Variadic template function (tmm_alloc_args) that wraps a memory allocation and subsequent placement new call .

From a users perspective:
`Foo* foo = tmm_alloc_args<Foo>()`  

The allocation functions takes the template parameters `<typename T, typename ...Args>` where Args is a parameter pack.  We can use this to wrap any constructor call, as long as we can  a string representation of the type `T` to pass to the trick memory manager.

The struct `TrickTypeToString` let's us get a string representation of the type through template specialization.  ICG will generate specializations for all types it knows about (except anything in Trick, std, or er7 namespaces).  

3. ICG Parses Constructors
I've added FunctionVisitor and FunctionDescription classes (modeled after FieldVisitor and FieldDescription). The visitor handles pulling some information out of `clang::Decl::CXXConstructor` and the description holds it.  This info is used by ICG to generate two files: `trick_type_to_string_ext.hh` and `constructors.json`.   As mentioned above, the TrickTypeToString<T> structures are used to convert types to strings.  The constructor information in constructors.json is used by convert_swig to generated addition functions in swig interface files.

```json
[
{
    "className": "AllocTestSimObject",
    "qualifiedClassName": "AllocTestSimObject",
    "constructors": [
    {
      "accessSpecifier": "public",
      "parameters": [
      ],
      "isDefault": true,
      "isCopy": false,
      "isMove": false
    }
    ]
}
]
```
If the constructor takes arguments, it'll have an array of type/name pairs. 

4.  SWIG Interface files
convert_swig has been updated with a new function `process_constructor_json` that uses the json data written by ICG to add some functionality.  Functions are added to the class using swigs %extend functionality.  One `alloc` function is generated for every constructor overload that exists.  

This is an example of the %extend functionality from test/SIM_tmm_alloc_args.  

```
%extend AllocTestWithArguments {
    static AllocTestWithArguments* alloc()  {
        return tmm_alloc_args<AllocTestWithArguments>();
    }
    static AllocTestWithArguments* alloc(int in_int, double in_double)  {
        return tmm_alloc_args<AllocTestWithArguments>(in_int, in_double);
    }
    static AllocTestWithArguments* alloc(int * in_int, double * in_double, std::string & a_name)  {
        return tmm_alloc_args<AllocTestWithArguments>(in_int, in_double, a_name);
    }
    static AllocTestWithArguments* alloc(int * arg_0, double & arg_1, std::string arg_2)  {
        return tmm_alloc_args<AllocTestWithArguments>(arg_0, arg_1, arg_2);
    }
}
```
SWIG will make those functions available in python for us, at this point a user can allocate using tmm_alloc_args from python.

`var = trick.AllocTestWithArguments.alloc(1, 5.0)`

To make it even more seamless I've modified the one of the `__init__` overrides in `swig_class_typedef.i` to call alloc using python's unpacking operator(*).  So now you can allocate using `trick.AllocTestWithArguments(1, 5.0)`.  

I've also added flags to ICG and convert swig so that all of the above features have to be explicitly enabled via:
```
TRICK_ICGFLAGS += -use_tmm_alloc_args
TRICK_CONVERT_SWIG_FLAGS += --tmm
```

This is still a work in progress, but if anyone spots any issues please let me know.


M-Herr:
Closing this for now.

PR #1799, opened 2024-10-24
Add calls for getting the total number of data recording groups and the data recording group pointer by its id number

hchen99:


excaliburtb:
this look perfect! thanks!

Issue #1802, opened 2024-11-12
Typo (ojbects)


simtheverse:
https://github.com/nasa/trick/blob/937712daa4ea2af52ae44033abcf78ca38e876a3/libexec/trick/pm/s_source.pm#L42

PR #1803, opened 2024-11-12
Update s_source.pm to fix typo, resolving #1802

simtheverse:
Update s_source.pm to fix typo, resolving #1802


sharmeye:
Thank you for bringing this to our attention. It is our policy that these sort of changes are handled internally and not via PR, but we'll make the changes you suggest next time we update our documentation. Thanks again for the suggestions.

PR #1810, opened 2024-12-03
TrickOps: Update send_hs.py to work with latest Trick output

### LINKED ISSUE #1807: TrickOps: Update send_hs.py to match latest send_hs output Trick produces
## Background

In the time since utility python module `share/trick/trickops/send_hs.py` was added to TrickOps (3/2023), Trick has moved on to a different output in the `send_hs` file that gets created, so this utility module is currently broken in the latest state of `master`. When I originally added `send_hs.py` it was not added with unit testing, so it's break was never discovered by Trick's CI.

## What this issue will address 

In this issue I will update `send_hs.py` to work with the latest output and add a unit test to cover it's usage, something I should have done in the first place!

## What we use it for

For the curious, we use this module to help us get `send_hs` data into a format we can more easily plot in our CI system. For example, here's one such plot showing us how our realtime performance ratio changes over time, where the x-axis is CI job number. It's a nice way for us to see if anything comes into our production branches that significantly affect realtime performance.

Screenshot 2024-11-20 at 3 23 07 PM

Each colored line represents one of our Orion scenarios we continually test in CI. Happy to provide more details on the plot if anyone in the community is interested in the details.

FYI @sharmeye @jmpenn @hchen99 for awareness

ddj116:
* Update send_hs.py to work with the new fields Trick's output produces
* Add unit tests and doctests for send_hs module
* Update TrickOps documentation to include info on send_hs usage
* Remove MonteCarloGenerationhelper.py's unecessary dependency on send_hs

Unrelated:

* Update pip when testing TrickOps per pip best practices

Closes #1807


ddj116:
Per discussion in the TrickLab this afternoon, the "check the source code for expected output" approach appears to be the best solution we can think of. Let me know if y'all have any other questions/concerns with the content of this PR. Thanks!

ddj116:
I realized I never flagged anyone on this! FYI @sharmeye @hchen99 @jmpenn @Pherring04

PR #1812, opened 2024-12-06
Don't override the CC and C++ environment variables

### LINKED ISSUE #1811: Don't Override User's CC and CPP Environment Variables
# Background

In the FSL we use `spack` to manage our software packages. So even though we're on Rocky 8, we can use GCC 12.2 via spack, for example. But when we tried to do this, we'd get a weird build error when building Trick:

```text
In file included from parse_format.c:3:
/usr/include/stdio.h:33:10: fatal error: stddef.h: No such file or directory
 #include <stddef.h>
```

So we just stuck with the system-installed GCC 8.5.

# Cause
Several of the makefiles in `trick_source/data_products/*` hard-code the following:

```makefile
CC = cc
CPP = c++
```

So the makefiles are stomping on the new compiler versions that we're trying to load via spack (or any other package manager). Because spack isn't actually updating `$PATH` for `cc`, it's just setting `$CC` and `$CPP`, this means that we're calling the system `/usr/bin/cc`  rather than the `/path/to/gcc12/cc` that we should be. We're mixing compiler versions for different parts of Trick, and eventually the build fails.

# Solution
We should be able to change those makefile lines to 

```makefile
CC ?= cc
CPP ?= c++
```

That way, if a user loads a new GCC via a package manager, Trick will use that GCC. Otherwise, we'll default to whatever's first on the user's path (probably the system-installed version).

FYI: @sharmeye  @ddj116 @hchen99. I have some local fixes but I don't think I have permission to push a new branch to Trick.

ninotarantino:
Fixes an issue where some makefiles override the user's CC and C++ environment variables. This can stomp on compiler versions loaded via a package manager like spack or conan.

Closes #1811

### DIFF
--- a/trick_sims/SIM_satellite/models/Satellite/graphics/makefile
+++ b/trick_sims/SIM_satellite/models/Satellite/graphics/makefile
@@ -1,5 +1,5 @@
-CPP      = g++
-CC       = gcc
+CPP     ?= g++
+CC      ?= gcc
 
 ifeq ($(shell uname), Darwin)
        GL_LIBS  = -framework GLUT -framework OpenGL
--- a/trick_sims/SIM_wheelbot/models/Battery/test/makefile
+++ b/trick_sims/SIM_wheelbot/models/Battery/test/makefile
@@ -1,8 +1,8 @@
 
 
 RM = rm -rf
-CC = cc
-CPP = c++
+CC ?= cc
+CPP ?= c++
 
 DECL_DIR = ..
 GTEST_DIR = ${HOME}/gtest-1.7.0
--- a/trick_sims/SIM_wheelbot/models/Control/test/makefile
+++ b/trick_sims/SIM_wheelbot/models/Control/test/makefile
@@ -1,6 +1,6 @@
 RM = rm -rf
-CC = cc
-CPP = c++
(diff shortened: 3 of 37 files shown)


ninotarantino:
@hchen99 This is ready for review.

ninotarantino:
I'm getting a really bizarre error when building some of the unit tests. Is it alright if we only modify the model directory makefiles and leave the test directory makefiles unchanged from master? This fix is really just to get Trick to build and I'm not too familiar with the test setup here. @hchen99

hchen99:
> I'm getting a really bizarre error when building some of the unit tests. Is it alright if we only modify the model directory makefiles and leave the test directory makefiles unchanged from master? This fix is really just to get Trick to build and I'm not too familiar with the test setup here. @hchen99

It's fine to just modify to get Trick to build.

ninotarantino:
Sounds good! I pushed an update. I ran the tests locally as well so we should be good now.

hchen99:
I did checkout your updates and were able to build trick and run all the tests. It seems strange that it can't pass our CI

sharmeye:
@ninotarantino can you close this one?  We merged @hchen99's branch that fixes this issue.

PR #1816, opened 2024-12-10
Drhdf5 fix

KatuunuXVI:
The Macros in the DRHDF5 file modify the size of the DRHDF5 in only this translation unit upon compilation. This can result in buffer overflow (encountered during tests with checkpoint loading).


sharmeye:
I think this will require more investigation to ensure that any proposed solution doesn't introduce unwanted behavior elsewhere in the system, and as such our team will investigate and resolve this internally.  Thank you for the information and valuable input you have provided.

PR #1819, opened 2024-12-12
1811 dont override users cc and cpp env for data products

hchen99:
Testing PR for PR #1812


sharmeye:
is this PR cursed?  it looks cursed

ninotarantino:
Looks cursed to me. What if instead of changing the `CC = cc` lines to `CC ?= cc` we just change them to `CC = gcc`? Surely that'll work. We're only really hitting this error because the `cc` on our system doesn't link to the proper `gcc`, but running `gcc` directly will work.

I definitely thought this PR would be simpler lol.

hchen99:
Yeah, unexpected. Do you still need to change CPP? Or you just need to change CC = cc to CC = gcc?

hchen99:
@ninotarantino Great, thanks for confirming!

Issue #1846, opened 2025-02-25
Add an optional datarecord sanity check


excaliburtb:
There is a problem with data products have a trick log files that contain the same variable logged more than once and at different rates. This results in anywhere from undefined behavior in plotting tools to just unintentionally incorrect plots. Trick could provide a verification check that can warn users if they have configured the sim to log variables in this conflicting way.

Consider adding a check either optionally or always but non-fatal to warn the user.

PR #1847, opened 2025-02-25
Add logging var multiple times check.

excaliburtb:
fixes https://github.com/nasa/trick/issues/1846


trick_source/sim_services/DataRecord/DataRecordDispatcher.cpp
```diff
@@ -69,8 +70,11 @@ void Trick::DRDWriterThread::dump( std::ostream & oss ) {
     Trick::ThreadBase::dump(oss) ;
 }
 
-Trick::DataRecordDispatcher::DataRecordDispatcher() : drd_writer_thread(drd_mutexes, groups) {
-    the_drd = this ;
+Trick::DataRecordDispatcher::DataRecordDispatcher()
+    : drd_writer_thread(drd_mutexes, groups),
+      verify_log_vars(true)
```
hchen99, line 75: default to false?

trick_source/sim_services/DataRecord/DataRecordDispatcher.cpp
```diff
@@ -333,6 +343,122 @@ int Trick::DataRecordDispatcher::init_groups() {
...
+    {
+        processMultipleVarLoggedCheck();
+    }
+
     return 0 ;
 }
 
+/**
+@details
+ * Checks the data logging configuration in Trick against certain error or warning conditions.
```
hchen99, line 357: Looks like 2 error checked: logged multiple times in the same group and in different groups. Don't see rate/total number of bytes checked. Assuming more code to be checked in as don't see any test code either?

hchen99:
Thanks and we'll discuss how we want to handle this PR.

trick_source/sim_services/DataRecord/DataRecordDispatcher.cpp
excaliburtb: oh sorry, that is a comment copy error where my project also had a warning for the expected number of bytes to log. for binary we pretty much know the sizes of the log file per second and am able to predict the resulting log file size according to the sim stop time. I'm not sure that's as useful for the general public

hchen99: Oh, I see. I was just reading the comment to understand what the function does, but the code seems to differ from what the comment describes. Didn’t mean to suggest one way or the other.

trick_source/sim_services/DataRecord/DataRecordDispatcher.cpp
excaliburtb: that's up to y'all but we're just printing warnings so I figure it should be defaulted true. otherwise, folks tend to not know it even exists.

trick_source/sim_services/DataRecord/DataRecordDispatcher.cpp
excaliburtb: removed the erroneous comments.

sharmeye:
I've taken this PR and made my own branch to work on to include levels of severity to the check.  Closing this PR in favor of PR #1942

PR #1853, opened 2025-02-27
Added interface file includes for C++ types supported by swig

M-Herr:
Update trick_swig.i swig interface file with more of the C++ Standard Library interface files provided by swig.


M-Herr:
Thanks for looking into this!  I've updated it to remove those redundant includes.

Issue #1898, opened 2025-06-03
Use "localhost" as hostname if no name given for TCP connections


alexlin0:
On some Mac laptops the hostname cannot be used to resolve an IP address.  Use "localhost" as hostname if the intended host is found to be the local machine.

alexlin0:
Not a Trick problem.

PR #1900, opened 2025-06-03
Use "localhost" as hostname if no name given for TCP connections #1898

alexlin0:
When connecting the sim control panel or other services using TCPClientListener, if no host is given, or the host is "localhost" or the host is equal to the name given back by the hostname routine, use "localhost" as the machine name.


alexlin0:
This changes the name shown in trick-sniffer and I don't like that.  Will need to find a different solution.

Issue #1907, opened 2025-06-04
Make Trick tests compatible with latest googletest 1.17 for Mac


alexlin0:
tests need to be compiled with c++17 to be compatible with googletest 1.17.

PR #1908, opened 2025-06-04
Make Trick tests compatible with latest googletest 1.17 #1907

alexlin0:
Changed a couple of test makefiles to specify c++17 instead of c++14.  Had to add common makefiles to the var_binary_parser makefile to get it to work outside of CI.


hchen99:
Related to issue #1907

PR #1910, opened 2025-06-05
Added or cases to optional job tags

Pherring04:
Made child and phase args optional.

### DIFF
--- a/libexec/trick/pm/parse_s_define.pm
+++ b/libexec/trick/pm/parse_s_define.pm
@@ -925,8 +925,8 @@ sub handle_sim_class_job($$$) {
       $ov_class , $ov_class_self , $sup_class_data, $tag, $job_call, $job_ret, $job_name, $args) = $in_job =~ /
          (?:
            \s*(?:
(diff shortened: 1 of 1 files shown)


timothyjbriscoe:
I can confirm this works for simulations without job tags.  But I did find the following...

On a simple spring mass simulation, if the job statement is: 

    P99 ("initialization" initialize();  

...then it build just fine as expected.  However, if I add a job tag and the statement is:

    {SPRING} P99 ("initialization" initialize(); 

  ...then the sim fails to build with the error: 

Writing    S_source.c
Writing    Makefile_swig
Running    ICG
/home/yoda/issue__parse_s_define/S_source.hh:836:84: error: use of undeclared identifier 'RING'
  836 |             job = this->add_job(0, 0, "initialization", NULL, 1, "initialize", "", RING) ;
      |                                                                                    ^
Writing    build/home/yoda/issue__parse_s_define/io_S_source.cpp
Writing    Makefile_io_src
Trick build was terminated due to error in user code!
makefile:108: build/Makefile_src: No such file or directory
make: *** [makefile:63: build/Makefile_io_src] Error 255

Issue #1912, opened 2025-06-06
JIT compile call should load resulting library


alexlin0:
When compiling JIT code, the code should automatically be loaded if successfully compiled.

PR #1913, opened 2025-06-06
JIT compile call should load resulting library #1912

alexlin0:
The compile command would track the resulting library, but didn't load the symbols.  This makes it impossible to do anything with the code.  Changed it so after successful compilation the library symbols are loaded.


sharmeye:
I'm surprised you didn't merge this one yourself

Issue #1914, opened 2025-06-09
FPEs don't work on Macs


alexlin0:
I suspect FPEs haven't been handled on the Mac in many years.

PR #1915, opened 2025-06-09
FPEs don't work on Macs #1914

alexlin0:
FPE signal handling has changed since the last time this code was touched, pre-2015.  Mac FPE signal handling works similar to linux now.  Was able to remove a lot of #if __APPLE__ sections of code. There is an Arm specific call on Apples to enable FPE, fesetenv. This is like linux specific feenableexcept.  I do not know what Macs on intel chips use, so I left that out, FPEs will still be broken on that platform combination.


hchen99:
It seems there is no supported way to enable floating-point exception traps at runtime on macOS with Intel chips. macOS Intel support will be added if a viable solution is found; otherwise, this limitation will be documented.

PR #1920, opened 2025-06-27
Fix DR_Changes_Step

### LINKED ISSUE #1919: DR_Changes_Step doesn't work
[DataRecordGroup::last_value](https://github.com/nasa/trick/blob/ac477deddaccf02a9d0f1bc7262a54879351026e/include/trick/DataRecordGroup.hh#L57) is never updated (except for [rec_buffer[0]](https://github.com/nasa/trick/blob/ac477deddaccf02a9d0f1bc7262a54879351026e/trick_source/sim_services/DataRecord/DataRecordGroup.cpp#L646), which is sim time). This results in log files with alternating rows of zeros (as `*last_value` is initialized to zero via [calloc](https://github.com/nasa/trick/blob/ac477deddaccf02a9d0f1bc7262a54879351026e/trick_source/sim_services/DataRecord/DataRecordGroup.cpp#L301)) and real values. This behavior appears to be the same in the [very first commit](https://github.com/nasa/trick/blob/f0c594f84106506625d94de11ed38ec3ffe710db/trick_source/sim_services/DataRecord/src/DataRecordGroup.cpp), which leads me to believe this feature has been broken for at least 10 years.

dbankieris:
Closes #1919

### DIFF
--- a/trick_source/sim_services/DataRecord/DataRecordGroup.cpp
+++ b/trick_source/sim_services/DataRecord/DataRecordGroup.cpp
@@ -687,21 +687,32 @@ int Trick::DataRecordGroup::data_record(double in_time) {
                 } else {
                    drb->curr_buffer += param_size ;
                 }
+                /**
+                 * While the typical idiom is something like:
+                 * 1. previous_value = current_value
+                 * 2. current_value = new_value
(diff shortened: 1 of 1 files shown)


dbankieris:
Alas, I cannot. But I know it to be true in my heart. I appreciate your indulgence.

PR #1941, opened 2025-08-14
Add extern C in swig autogenerated cpp files for c includes

mrjaperez:
The autogenerated *_py.cpp files do not contain the extern C guards around the C include files, causing unresolved symbols during linking.

### DIFF
--- a/libexec/trick/convert_swig
+++ b/libexec/trick/convert_swig
@@ -328,7 +328,7 @@ sub process_file() {
 #include <cstddef>
 \%}\n" ;
 
-        if ($f =~ /.hh/) { #Check if CPP file
+        if ($f =~ /.hh/ || $f =~ /.hpp/) { #Check if CPP file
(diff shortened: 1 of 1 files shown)


mrjaperez:
Correct, will get undefined references when linking the C symbols in the C++ swig generated code.

mrjaperez:
When you #include a C header in C++ you need to do extern C to get the correct linkage. This change adds that to the swig C++ auto-generated code.

hchen99:
If you have C functions defined in C header that need to be accessed by C++ code, add 
`#ifdef __cplusplus
extern "C" {
...` in you C header.

mrjaperez:
sorry, moved to quick on me.

My change is exactly what it is suggested above, but it is doing it in the correct place. Please take a closer look at the change and where it is being added. Don't need to have users add all that, it is done automatically and done it the correct place.

Issue #1944, opened 2025-08-21
Add python virtual environment path to Input processor search path


alexlin0:
If start a sim in a python virtual environment, that virtual environment is not passed down to the embedded input processor.  Packages installed in the virtual env are not available to the sim.

PR #1945, opened 2025-08-21
Add python virtual environment path to Input processor search path #1944

alexlin0:
Added an if check before in the python code run before the input file and test if the VIRTUAL_ENV environment variable is set.  If it is, add the path ${VIRTUAL_ENV}/python_<version>/site-packages to sys.path list.


sharmeye:
I don't think this will work for Conda virtual envs, but it won't hurt either.  For Conda envs, I think it would need to look for CONDA_DEFAULT_ENV and CONDA_PREFIX instead of VIRTUAL_ENV.  That's a later issue to address, this one gets to go in now.

PR #1966, opened 2025-09-26
Introduce the option to include decimal comment lines for hexfloat values in the checkpoint file.

hchen99:
Introduce the option to include decimal comment lines for hexfloat values in the checkpoint file for readability if preferred. By default, no comment lines.

`trick.TMM_hexfloat_decimal_comment_checkpoint(True)`


excaliburtb:
good great! good to go from my perspective

PR #1972, opened 2025-09-30
Updated to write string literal format for c-string for checkpoint.

hchen99:
Updated to write string literal format for c-string for checkpoint.


hchen99:
This PR is cursed. Will open another one when it's ready.

PR #1988, opened 2025-10-29
Optimization directory and LDLt factorization

narvindk:
Added primal-dual interior point algorithm in a new optimization directory in trick_utils and dLDLt_solver in trick_utils/math.


sharmeye:
Making a small update and merging via PR #2053

PR #2006, opened 2025-12-09
Enforce cpp17

sharmeye:
Added cpp17 to top-level Makefile, and removed cpp11 flags from every Makefile it was explicitly put in.


test/SIM_stls/models/STLCheckpoint.hh
```diff
@@ -124,5 +127,14 @@ class STLCheckpoint {
...
 } ;
 
+class StringViewCheck
+{
+    public:
+        std::string_view    stringViewSample;
+
+        void string_view_print();
+
+};
```
brendan-nasa: why'd you add this class?

brendan-nasa:
T_T

https://github.com/nasa/trick/blob/027e36e3b4d5dc7936c266d57515dd6f7cad7080/test/SIM_stls/S_overrides.mk#L5

PR #2009, opened 2025-12-10
Add typemap for command_line_args_get_argv

### LINKED ISSUE #2008: `char**` breaks python input processor
Noting here that our Ramtares CI system that forward-tests our workflow with the bleeding edge of Trick's `origin/master` is consistently failing with this type of error at sim startup:

```
  File "RUN_AOA/input.py", line 8, in <module>
    exec(open("Modified_data/input_files/default_AOA.py").read())
  File "<string>", line 8, in <module>
  File "<string>", line 98, in __init__
IndexError: list index out of range
|L   0|2025/12/04,22:02:15.849387|blade17-fsl.fsl.jsc.nasa.gov| |T 0|0.000000| [00m
SIMULATION TERMINATED IN
  PROCESS: 0
  ROUTINE: IPPython.cpp:148
  DIAGNOSTIC: Input Processor error
```
We had a successful test at b6061e76d650e3c57bbdccf30c727252084b970f which is the direct parent of 36743883e153ae364f48596942e3cda3d3e5dca7 so there appears to be something about this PR that is causing this. 

This isn't currently breaking us as our production branches are hanging out at much older 16828a9000c77f013a6edb0e324e284ecfa18ae4 and will be for some time, just posting here in case others in the Trick community are seeing something similar. Will update this thread with info as we dig in.

_Originally posted by @ddj116 in https://github.com/nasa/trick/issues/1980#issuecomment-3617864829_

brendan-nasa:
Closes #2008

### DIFF
--- a/trick_source/sim_services/CommandLineArguments/test/CommandLineArguments_test.cpp
+++ b/trick_source/sim_services/CommandLineArguments/test/CommandLineArguments_test.cpp
@@ -2,51 +2,51 @@
 #include "trick/command_line_protos.h"
 
 #include <cstdio>
+#include <cstdlib>
 #include <string>
 #include <sys/stat.h>
 #include <unistd.h>
+#include <utility>
 #include <vector>
 
 #include <gtest/gtest.h>
 
 // Helper for directory cleanup
-class TempDirectory
-{
+class TempDirectory {
     std::string path_;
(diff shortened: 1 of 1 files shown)


hchen99:
Thinking of adding tests for command_line_args_get_argc and command_line_args_get_argv in the CommandLineArguments test folder, and also adding tests to input.py under SIM_test_output_dir.

PR #2017, opened 2025-12-17
Fixed potential logging offset from restart

excaliburtb:
from v25.0.0, if a checkpoint was cut on a sim cycle offset of a dr group's cycle, then the offset would be introduced on restart. This PR fixes that issue and updates a test to demonstrate it.


hchen99:
Update README.md for the test?

PR #2021, opened 2025-12-29
Add precision specification to swig_double

### LINKED ISSUE #2019: swig_double print loses precision
the repr function for the swig_double helper class uses a simple `%g` for the formatter. This isn't the same as what is in the PythonPrint.cpp class when printing from memory-managed doubles. The issue is the loss of precision in the print.

For example: the original trick swig_double for the time = `4732992000.0` would print as `4.73299e+09` on Mac 

My use case is a JEOD sim setting the stop time to 4732992000, then using the exec_get_terminate_time() in a formatted string to do something with that time and saw this error.

```python
trick.stop(4732992000)
print(f'end_time = {trick.exec_get_terminate_time()}')
```

end_time would print as 4.73299e+09 losing an entire 2000 seconds

excaliburtb:
Fixes #2019

### DIFF
--- a/test/SIM_test_ip/RUN_test/unit_test.py
+++ b/test/SIM_test_ip/RUN_test/unit_test.py
@@ -23,6 +23,8 @@ def main():
 
     test_suite = "double"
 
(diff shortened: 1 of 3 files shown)


excaliburtb:
Added test. Should be good to go.

PR #2077, opened 2026-03-23
Applied ICG filtering based on TRICK_EXCLUDE & TRICK_ICG_EXCLUDE & TRICk_SYSTEM_EXCLUDE across the board.

### LINKED ISSUE #2065: TRICK_EXCLUDE not working properly
Using Trick test sim version 19.8.0-beta:
trick/test/SIM_exclusion_mechanisms

If you add this comment block in models/Exclude.hh lines 12-14:

Image

It gives me ICG errors even though the S_overrides.mk has:
TRICK_EXCLUDE += :$(MODELS)/Exclude.hh

hchen99:
Applied ICG filtering based on TRICK_EXCLUDE & TRICK_ICG_EXCLUDE & TRICk_SYSTEM_EXCLUDE across the board.


trick_source/codegen/Interface_Code_Gen/Utilities.cpp
brendan-nasa, line 1: why `and` instead of `&&` ?

hchen99: because they're functionally identical in c++ and also probably just did some python coding so for python friendly :-)

PR #2092, opened 2026-04-15
Update Mtv to correctly use condition_list and action_list

### LINKED ISSUE #2020: MTV failing to launch. Variable Server cant find cond[] and act[]
Attempting to open the MTV gives this error: 

Image

mtv list variables:

Image

It appears that cond[0] and act[0] were updated to condition_list[0][0] and action_list[0][0] but the mtv was not updated to match. Here are the variables that were changed https://github.com/nasa/trick/blame/master/include/trick/IPPythonEvent.hh#L114-L116 

It appears the MTV java is still using cond[] and act[] https://github.com/nasa/trick/blob/master/trick_source/java/src/main/java/trick/mtv/MtvApp.java
 even though the trick event class has been updates

I could be missing something else but my mtv is not opening and I believe it is because the mtv java has not been updates to match the new mtv variables.
Image

brendan-nasa:
test doesn't run the gui directly but opens a socket and sends the same fields that Mtv uses

### DIFF
--- a/test_sims.yml
+++ b/test_sims.yml
@@ -1,8 +1,6 @@
 # Compile only sims
 SIM_alloc_test:
   path: test/SIM_alloc_test
-SIM_alloc_test:
-  path: test/SIM_alloc_test
 SIM_anon_enum:
   path: test/SIM_anon_enum
 SIM_default_member_initializer:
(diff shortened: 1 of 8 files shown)


test_sims.yml
```diff
@@ -1,6 +1,4 @@
 # Compile only sims
-SIM_alloc_test:
-  path: test/SIM_alloc_test
 SIM_alloc_test:
   path: test/SIM_alloc_test
```
brendan-nasa, line 3: why was this in here twice?

trick_source/java/src/main/java/trick/mtv/MtvApp.java
```diff
@@ -260,34 +256,34 @@ protected void initialize(String[] args) {
...
                         mtv_var_count += 4;
                     }
-                    for (int jj=0; jj < mtv_events.get(ii).act_count; jj++) {
-                        event_vars += "trick.var_add(\"trick_ip.mtv.mtv_list[" +ii+ "][0].act[" +jj+ "].enabled\"); ";
-                        event_vars += "trick.var_add(\"trick_ip.mtv.mtv_list[" +ii+ "][0].act[" +jj+ "].ran_time\"); ";
-                        event_vars += "trick.var_add(\"trick_ip.mtv.mtv_list[" +ii+ "][0].act[" +jj+ "].ran_count\"); ";
+                    for (int jj = 0; jj < mtv_events.get(ii).act_count; jj++) {
+                        event_vars += "trick.var_add(\"trick_ip.mtv.mtv_list[" + ii + "][0].action_list[" + jj + "].enabled\"); ";
+                        event_vars += "trick.var_add(\"trick_ip.mtv.mtv_list[" + ii + "][0].action_list[" + jj + "].ran_time\"); ";
+                        event_vars += "trick.var_add(\"trick_ip.mtv.mtv_list[" + ii + "][0].action_list[" + jj + "].ran_count\"); ";
```
brendan-nasa, line 286: these are the important lines in MtvApp

PR #2099, opened 2026-04-21
Fixes to MSSharedMem class

### LINKED ISSUE #2098: MSSharedMem doesn't support more than 1 slave
MSSharedMem has a couple of issues:

1. the memory layout of the data in the shared memory block is not explicitly padded to avoid issues with a slave that compiles differently.
2. the proj_id within the tsm library is a static int that increments regardless of user input. so a second slave cannot get the same shm key.

The result of the above is that a 32-bit slave is unlikely to connect correctly even as the sole slave. 

If more than 1 MSSharedMem slave is used, the second slave can generate a key and both master and slave think they were successful but they will not communicate with each other properly

excaliburtb:
- Updated MSSharedMemData struct to eliminate compiler padding due to alignment. 

- Updated TSM library to dole out proj_id values only if unset by the user and to track them on a per key_file basis.


include/trick/tsm.h
```diff
@@ -34,9 +34,12 @@ typedef struct {
     int size;                          /* -- Size of shared memory segment */
     void* addr;                        /* ** Address of shared memory segment */
 
-    char key_file[256];                 /* -- An existing file name to use in generating key (ftok) for shared memory */
-    int  key_id;                       /* ** Id value to use in generating key value (ftok) for shared memory */
+    char key_file[256];                /* -- An existing file name to use in generating key, in combination with proj_id, (ftok) for shared memory */
+    int  proj_id;                      /* -- A value to use, in combination with the key_file, (ftok) for shared memory. Use 1..255 to create 255 different keys */
+    int  key_id;                       /* ** Id value to use,  in generating key value (ftok) for shared memory. . */
```
hchen99: Looks like key_id is an orphaned field

trick_sims/SIM_Ball++_L1/local_trick.sh
brendan-nasa, line 1: was this supposed to make it in?

include/trick/MSSharedMem.hh
```diff
@@ -61,12 +63,12 @@ namespace Trick {
 
     /** The data to read/write between the master and slave in shared memory.\n */
     typedef struct {
-        pid_t master_pid ;                      /**< trick_units(--) */
-        MSQ_DECLARE (master_time, long long)
-        MSQ_DECLARE (master_command, MS_SIM_COMMAND)
-        MSQ_DECLARE (slave_command, MS_SIM_COMMAND)
+        uint64_t master_pid ;                      /**< trick_units(--) */
```
hchen99: int64_t?

trick_sims/SIM_Ball++_L1/RUN_shm_slaves/slave2_shm2.py
```diff
@@ -0,0 +1,18 @@
...
+#this will be turned off if the sim is a slave
+trick.real_time_enable()
+
+trick.exec_set_software_frame(0.1)
+trick.exec_set_freeze_frame(2.00)
+#trick.exec_set_freeze_command(True)
+
+new_connection = trick.MSSharedMem()
+new_connection.tsm_dev.key_file = trick.get_trick_env("TRICK_HOME") + "/trick_sims/SIM_Ball++_L1/RUN_shm_slaves/slave2_shm2.py"
+new_connection.tsm_dev.proj_id = 5
```
hchen99, line 13: if not setting proj_id to 5, the 2nd shm slave also gets proj_id as 1 as the 1st shm slave... should proj_id for the 2nd shm slave be 2 if both slaves are managed ?

trick_sims/SIM_Ball++_L1/RUN_test/master.py
```diff
@@ -25,8 +57,9 @@
 #trick.freeze(5.999999)
 #trick.add_read(4.9999, "trick.freeze()")
 
-print "Number of slaves...", trick_master_slave.master.num_slaves
-print "The 1st slave sim path...", trick_master_slave.master.slaves[0].sim_path
+print(f"Number of slaves...{trick_master_slave.master.num_slaves}")
+for ii in range(trick_master_slave.master.num_slaves):
+   print(f"Slave {ii} sim path...{trick_master_slave.master.slaves[ii].sim_path}/{trick_master_slave.master.slaves[ii].run_input_file}")
```
hchen99, line 62: thanks for updating for python 3

trick_source/trick_utils/shm/src/tsm_init.c
```diff
@@ -54,17 +64,49 @@ int tsm_init(TSMDevice * shm_device)
...
+        if(foundPair == NULL)
+        {
+            g_TrickShmPairs = realloc(g_TrickShmPairs, sizeof(TRICK_SHM_PAIRS)*(g_numTrickShmPairs+1));
+            memset(&g_TrickShmPairs[g_numTrickShmPairs], 0, sizeof(TRICK_SHM_PAIRS));
+            strncpy(g_TrickShmPairs[g_numTrickShmPairs].key_file, shm_device->key_file, sizeof(shm_device->key_file));
+            g_TrickShmPairs[g_numTrickShmPairs].proj_id = 1;
+            shm_device->proj_id = 1;
+            ++g_numTrickShmPairs;
+        } else {            
+            shm_device->proj_id = ++foundPair->proj_id;
```
hchen99: need to check against 255?

trick_source/trick_utils/shm/src/tsm_init.c
```diff
@@ -54,17 +64,49 @@ int tsm_init(TSMDevice * shm_device)
...
+            TRICK_SHM_PAIRS * testPair = &g_TrickShmPairs[ii];
+            if(strncmp(testPair->key_file, shm_device->key_file, sizeof(testPair->key_file)) == 0)
+            {
+                foundPair = testPair;
+                break;                
+            }
+        }
+        if(foundPair == NULL)
+        {
+            g_TrickShmPairs = realloc(g_TrickShmPairs, sizeof(TRICK_SHM_PAIRS)*(g_numTrickShmPairs+1));
```
hchen99: need to free the memory for ralloc?

trick_sims/SIM_Ball++_L1/local_trick.sh
excaliburtb: yes, though it's just bash so i should change the extension to sh. it's pretty much required to have a "trick" environment for master/slave (annoyingly)

trick_sims/SIM_Ball++_L1/RUN_shm_slaves/slave2_shm2.py
excaliburtb: i'm testing that proj_id can be set and utilized properly. and it needs to match between master and slave which is why master also has it set to 5

trick_source/trick_utils/shm/src/tsm_init.c
excaliburtb: realloc frees or at least invalidates the original pointer passed in and does not require a free. in fact, it's undefined behavior

include/trick/MSSharedMem.hh
excaliburtb: i think i went unsigned out of habit. will change..

trick_sims/SIM_Ball++_L1/RUN_shm_slaves/slave2_shm2.py
hchen99: Was just wondering if not setting proj_id for shm slave2, whether the proj_id for shm slave2 should be set to 2 instead of 1. Thought proj_id wasn't incremented if proj_id was not set.

Tested again and I believe proj_id was set as intended:

- if not setting proj_id as tsm_dev.proj_id = 5 for shm slave2 
  - if using the same key_file for shm slave1 & slave2, slave2 did get proj_id 2 while slave1 got proj_id 1
    - if same key_file, proj_id gets incremented
  - if using different key_file for shm slave1 & slave2, slave1 got proj_id 1 and slave2 also got proj_id 1 because the the key_file is different, therefore the generated key is different due to pair of key_file and proj_id
    - if different key_file, proj_id can be the same for different slaves

Meanwhile, your test did show proj_id = 5 was honored.

trick_source/trick_utils/shm/src/tsm_init.c
hchen99: while realloc frees old memory block and returns a new one, the new/final one still needs free, is that correct? just trying to understand... I guess since they are static globals that live for the entire process and the OS reclaims the memory on exit anyway, so ok to skip the cleanup ;-)

excaliburtb: yes, the new pointer will need to be freed but there's not a good opportunity to do that since it's a global pointer as you already stated. if there's an idea how to do that, I'm up for any suggestions but being in C i don't see how, lol.

include/trick/tsm.h
excaliburtb: maybe that was intended to do what i ended up doing but never got there..

trick_source/trick_utils/shm/src/tsm_init.c
excaliburtb: will add...

trick_source/trick_utils/shm/src/tsm_init.c
hchen99: register the cleanup call with atexit on first allocation. but only works when the process exits normally but sim can be terminated via signals, so only atexit might not always get called ;-)

include/trick/MSSharedMem.hh
excaliburtb: fixed

trick_source/trick_utils/shm/src/tsm_init.c
excaliburtb: registered a cleanup with atexit. i confused it with a signal handler thinking only one function could register with atexit. easy fix

trick_source/trick_utils/shm/src/tsm_init.c
excaliburtb: added

trick_sims/SIM_Ball++_L1/local_trick.sh
excaliburtb: updated

include/trick/tsm.h
excaliburtb: removed

trick_source/trick_utils/shm/src/tsm_init.c
```diff
@@ -32,13 +45,39 @@ key_t my_ftok(const char* path, int proj_id) {
 
 }
 
+static void cleanup_trickshm_pairs(void)
+{
+    if (g_TrickShmPairs != NULL)
```
hchen99: Learned this morning that `if (g_TrickShmPairs != NULL)` is not necessary since free(NULL) is no op and always safe. And it might not clean up g_numTrickShmPairs if g_TrickShmPairs is NULL for some reason globals are out of sync ;-)

trick_source/trick_utils/shm/src/tsm_init.c
hchen99: How about validating the next value first before setting it to the cached pair so it wouldn't keep incrementing for the same file after 255?

trick_source/trick_utils/shm/src/tsm_init.c
```diff
@@ -54,17 +93,60 @@ int tsm_init(TSMDevice * shm_device)
...
+            if (strncmp(testPair->key_file, shm_device->key_file, sizeof(testPair->key_file)) == 0)
+            {
+                foundPair = testPair;
+                break;
+            }
+        }
+        if (foundPair == NULL)
+        {
+            g_TrickShmPairs = realloc(g_TrickShmPairs, sizeof(TRICK_SHM_PAIRS) * (g_numTrickShmPairs + 1));
+            memset(&g_TrickShmPairs[g_numTrickShmPairs], 0, sizeof(TRICK_SHM_PAIRS));
```
hchen99, line 125: If realloc fails, would g_TrickShmPairs lose the old allocation and deference a null pointer in memset? Maybe add a check?

trick_source/trick_utils/shm/src/tsm_init.c
```diff
@@ -54,17 +93,60 @@ int tsm_init(TSMDevice * shm_device)
...
+            strncpy(g_TrickShmPairs[g_numTrickShmPairs].key_file, shm_device->key_file, sizeof(shm_device->key_file));
```
hchen99, line 126: Wondering if need a forced termination?

trick_source/trick_utils/shm/src/tsm_init.c
excaliburtb: yes, but trick isn't checking if malloc fails globally. i.e. you have to pretty much run out of memory for it to fail, which is the same reason realloc might fail. feel free to add a check though

trick_source/trick_utils/shm/src/tsm_init.c
excaliburtb: i don't recommend that in a library which the shm directory seems to be treated. a return code and maybe a print at worst is the recommended practice

trick_source/trick_utils/shm/src/tsm_init.c
hchen99: yeah, trick isn't checking for the most part...not picking on you ...updated ;-)

trick_source/trick_utils/shm/src/tsm_init.c
hchen99: updated

PR #2100, opened 2026-04-21
Added special handling for er7_utils::Integration enums previously unrecognized by SWIG.

### LINKED ISSUE #692: Error when setting member of type er7_utils::Integration::Technique from Python context
S_define
```c++
#include "sim_objects/default_trick_sys.sm"
##include "trick/Integrator.hh"

class Sandbox : public Trick::SimObject {
    public:
    er7_utils::Integration::Technique technique;
};

Sandbox sandbox;
```

input.py
```python
sandbox.technique = trick.Integration.SymplecticEuler
```

Result
```python
  File "RUN_test/input.py", line 1, in <module>
    sandbox.technique = trick.Integration.SymplecticEuler
TypeError: in method 'Sandbox_technique_set', argument 2 of type 'er7_utils::Integration::Technique'
```

ccsullivan2:
Fixes #692

### DIFF
--- a/include/trick/swig/swig_int_typemap.i
+++ b/include/trick/swig/swig_int_typemap.i
@@ -93,6 +93,42 @@
     $result = SWIG_NewPointerObj(SWIG_as_voidptr(t), SWIG_TypeQuery("_p_swig_int"), SWIG_POINTER_OWN);
 }
 
+#ifdef USE_ER7_UTILS_INTEGRATORS
+// Special handling for er7_utils::Integration enums.
+// These enums are nested inside a class and are not recognized by SWIG
+// as standard enums in generated modules, so provide explicit
+// scalar conversion support.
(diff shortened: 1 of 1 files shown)


hchen99:
Have you tried to add something like following to your S_define (using the example in #692) in order to copy all typemaps from a known type to new ones?
```C++
%header{
##if defined(SWIG) && defined(USE_ER7_UTILS_INTEGRATORS)
    %apply int { er7_utils::Integration::Technique };
    %apply int { er7_utils::Integration::ODEProblemType };
##endif
%}
class Sandbox : public Trick::SimObject {
```

sharmeye:
@hchen99's suggestion is a more compact way of achieving this, and we might put her implementation in the appropriate area in the future.  For now I suggest you take her advice.

PR #2107, opened 2026-04-29
Checkpoint in freeze only

sharmeye:
Implement constraint to allow checkpoints only in freeze mode to preserve homogeneity of data.  Also implementing auto-freezing and auto-unfreezing the simulation when a checkpoint is attempted in RUN mode.  Credit to @Mrockwell2


brendan-nasa:
this MR adds a circular dependency: `Executive <=> CheckPointRestart`

trick_source/sim_services/CheckPointRestart/CheckPointRestart.cpp
```diff
@@ -257,12 +284,20 @@ int Trick::CheckPointRestart::write_checkpoint() {
...
+            file_name = chk_name_stream.str();
+        } else {
+            file_name = chkpnt_names[curr_time];
+            chkpnt_names.erase(curr_time);
+        }
 
-        checkpoint( chk_name_stream.str() );
+        checkpoint( file_name );
 
+		the_exec->run();
```
brendan-nasa: i think this unfreezes after checkpoint even when it wasn't an auto-freeze

also: indentation

hchen99:
> this MR adds a circular dependency: `Executive <=> CheckPointRestart`

It doesn't look like there is circular a dependency at the header/interface level even though CheckPointRestart.cpp includes Executive.hh and Executive_freeze_loop.cpp includes CheckPointRestart.hh. Each .cpp compiles independently with no include cycle and both compile into the same libtrick.a.

trick_source/sim_services/CheckPointRestart/CheckPointRestart.cpp
hchen99: In this branch, only scheduled checkpoint such as calling checkpoint(time, ... ) sets freeze at the specified time, then freeze_loop calls write_checkpoint at the scheduled time which sets to Run back at end. So it's safe.

When calling checkpoint without a time (not scheduled):
- If in Freeze, doesn't go through the route as stated above
- If not in Freeze, warning and no checkpoint done

The auto_freeze flag is only used for checkpoint loading. And we'll discuss if we want to only allow loading when in Freeze mode. This branch uses auto_freeze flag to unfreeze when loading is called in Run mode.

trick_source/sim_services/CheckPointRestart/CheckPointRestart.cpp
```diff
@@ -174,6 +184,19 @@ int Trick::CheckPointRestart::do_checkpoint(std::string file_name, bool print_st
 
     JobData * curr_job ;
     pid_t pid;
+    SIM_MODE mode;
+
+    mode = the_exec->get_mode();
+
+    if (mode == Run) {
```
hchen99: Should it be `mode != Freeze`, otherwise if in Initialization mode, can still do checkpoint?

trick_source/sim_services/CheckPointRestart/CheckPointRestart.cpp
hchen99: However though, if input file has both schedule freeze and checkpoint for the same time, after checkpoint, it'd go to unfreeze automatically which might not be as expected. So needs to check if there is scheduled freeze before unfreeze.

trick_source/sim_services/CheckPointRestart/CheckPointRestart.cpp
```diff
@@ -304,6 +357,20 @@ int Trick::CheckPointRestart::safestore_checkpoint() {
...
 
 void Trick::CheckPointRestart::load_checkpoint(std::string file_name) {
+    SIM_MODE mode = the_exec->get_mode();
+
+    if (mode == Run) {
+        std::string msg_format  = "WARNING: Loading a checkpoint in 'Run Mode' may cause non time-homogeneous data. ";
+                    msg_format += "Current Mode: %s (%d)\n";
+        
+        message_publish(MSG_WARNING, msg_format.c_str(),
+                        file_name.c_str(), simModeCharString(mode), mode);
```
brendan-nasa: too many parameters for the format string

trick_source/sim_services/Executive/Executive_freeze_loop.cpp
```diff
@@ -32,6 +35,10 @@ int Trick::Executive::freeze_loop() {
     }
 
     message_publish(MSG_INFO, "Freeze ON. Simulation time holding at %f seconds.\n" , get_sim_time()) ;
+    
+    if (!the_cpr->checkpoint_times.empty()) {
+    	the_cpr->write_checkpoint();
```
brendan-nasa: tab character

trick_source/sim_services/CheckPointRestart/CheckPointRestart.cpp
```diff
@@ -304,6 +357,20 @@ int Trick::CheckPointRestart::safestore_checkpoint() {
...
+        // If in RUN mode, this will freeze the simulation and notify the code to unfreeze later.
+	// To forbid loading a checkpoint in RUN mode, remove the following two lines and the second to last line in load_checkpoint_job()
```
brendan-nasa: \t

PR #2117, opened 2026-05-12
Completed std::wstring support.

### LINKED ISSUE #2094: Trick does not support std::wstring
Even though Trick supports wide characters (i.e. wchar_t), it appears Trick does not support the std::wstring type.  However, the TRICK_TYPE enumeration seems to imply std::wstring is supported because there is a TRICK_WSTRING entry. Native Trick support for std::wstring will be needed for TrickHLA since std::wstring is being used and we will need to checkpoint this data type.

Using the Trick MemoryManager, I get these errors when using a std::wstring.

MemoryManager:ERROR:Couldn't find function "io_src_sizeof_std__wstring()". dlerror= dlsym(RTLD_DEFAULT, io_src_sizeof_std__wstring): symbol not found

MemoryManager:ERROR:size of type "std::wstring" not found.

MemoryManager:ERROR:get_type_attributes failed for type: std::wstring

Process terminated by signal SIGSEGV

hchen99:
Tested following for this ticket for both std::wstring and std::string:

- std::wstring _wstring, std::wstring[2] vec2_wstring, and std::wstring* ptr_wstring
  - In input file: 
    - xxx._wstring = "Hello"
    - xxx.vec2_wstring = ["Hello", "World"]
    - xxx.ptr_wstring = trick.TMM_declare_var_1d("std::wstring", 5)
- checkpoint
- TV

### DIFF
--- a/trick_source/sim_services/VariableServer/VariableReference.cpp
+++ b/trick_source/sim_services/VariableServer/VariableReference.cpp
@@ -421,7 +421,7 @@ int Trick::VariableReference::stageValue(bool validate_address) {
         if (_address == NULL) {
             _size = 0 ;
         } else {
-            _size = wcslen((wchar_t *)_address) * sizeof(wchar_t);
+            _size = (wcslen((wchar_t *)_address) + 1) * sizeof(wchar_t);
         }
     }
     if(_address != NULL) {
--- a/trick_source/sim_services/VariableServer/VariableReference.cpp
+++ b/trick_source/sim_services/VariableServer/VariableReference.cpp
@@ -395,6 +395,14 @@ int Trick::VariableReference::stageValue(bool validate_address) {
         _address = (void *)(str_ptr->c_str()) ;
     }
 
+    // if this variable is a wstring we need to get the raw wide character string out of it.
+    if ((_trick_type == TRICK_WSTRING) && !_deref)
+    {
+        std::wstring *wstr_ptr = (std::wstring *)_var_info->address;
+        // Get a pointer to the internal wide character array
+        _address = (void *)(wstr_ptr->c_str());
+    }
+
     // if this variable itself is a pointer, dereference it
     if ( _deref ) {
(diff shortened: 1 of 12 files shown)


trick_source/sim_services/VariableServer/VariableReference.cpp
brendan-nasa: does this need to make room for null terminator?

hchen99: Yes, good catch! Even though writeValueAscii would still find a null terminator, but writeValueBinary uses exact size bytes, also, it'd be consistent with TRICK_STRING with room for null terminator.

trick_source/trick_swig/swig_ref.cpp
```diff
@@ -1063,6 +1063,10 @@ PyObject * swig_ref::__getitem__(int ii) {
                 t->units = temp_ref.attr->units ;
                 ret = SWIG_NewPointerObj(SWIG_as_voidptr(t), SWIG_TypeQuery("_p_swig_int"), SWIG_POINTER_OWN);
             } break ;
+            case TRICK_STRING: {
+                std::string * str_ptr = (std::string *)temp_ref.address ;
```
brendan-nasa, line 1067: I was _just_ about to ask "what about `TRICK_STRING`?" :)

how was this working before though?

hchen99: Probably mostly just use basic string variable. A plain scalar assignment doesn't create a swig_ref. The support is provided by swig built-in std_string.i. Before, in input file, while `xxx._string = "_string...Hello World"` works, but something like `xxx.vec3_string[0] = "vec3_string...Hello"` or `xxx.vec3_string = ["str1", "str2", "str3"]` would not work also `xxx.ptr_string = trick.TMM_declare_var_1d("std::string", 5)` doesn't work either.

PR #2123, opened 2026-05-14
ci: declare workflow-level `contents: read` on the 11 remaining workflows

arpitjain099:
Pins the default `GITHUB_TOKEN` to `contents: read` on the 11 workflows in `.github/workflows/` that don't actually need any write scope:

- `alt_linux_distros.yml`, `docker.yml`, `test_32_oracle.yml`, `test_linux.yml`, `test_linux_py2.yml`, `test_macos.yml`, `trickops.yml`: build and test workflows, no GitHub API mutation.
- `report_alt_linux_distros.yml`, `report_linux.yml`, `report_linux_py2.yml`: status-report workflows that read CI artifacts and post no comments.
- `code_coverage.yml`: uploads to Coveralls via `coverallsapp/github-action`. The `github-token` passed to that action is used to identify with coveralls.io, not for any GitHub API mutation, so `contents: read` is sufficient.

## Why

CVE-2025-30066 (March 2025 `tj-actions/changed-files` supply-chain compromise) exfiltrated `GITHUB_TOKEN` from workflow logs and the leaked token retained whatever scope was issued at the workflow level. Pinning per workflow caps that runtime authority irrespective of the repo or org default, gives drift protection if the default ever widens, and is credited per-file by the OpenSSF Scorecard `Token-Permissions` check.

YAML validated locally with `yaml.safe_load` on each touched file.

### DIFF
--- a/.github/workflows/alt_linux_distros.yml
+++ b/.github/workflows/alt_linux_distros.yml
@@ -5,6 +5,9 @@ on:
     - cron: "0 0 * * 0"
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   build:
     strategy:
--- a/.github/workflows/code_coverage.yml
+++ b/.github/workflows/code_coverage.yml
@@ -26,6 +26,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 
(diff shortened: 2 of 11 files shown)


sharmeye:
This was independently worked in another current PR, thank you for your submission.

PR #2131, opened 2026-05-24
ci: declare workflow-level `contents: read` on test_linux, test_macos, code_coverage

arpitjain099:
Three CI workflows just run language tests / coverage. No GitHub API writes from the workflows themselves, so workflow-level `contents: read` is the appropriate cap for the default `GITHUB_TOKEN`.

Same post-CVE-2025-30066 (`tj-actions/changed-files`) hardening shape. YAML validated locally.


sharmeye:
This was independently worked in another current PR, thank you for your submission.

PR #2133, opened 2026-05-28
Handle disabled variable server

brendan-nasa:


hchen99:
Can we also add when variable server status showing neither yes nor no, grey out right-click?
I take it back because the reason for blank is the older version of trick doesn't send the variable server enabled status

PR #2149, opened 2026-06-16
Fix duplicate struct cast_as generated in .i file.

hchen99:
Fix the linking issue due to duplicate struct/class cast_as generated in .i files (castAs in _py.cpp) if the same typedef declaration in different header files. The linker would see duplicates if different headers share the same tyepdef if with program scope. 
Before PR #2111, cast_as never got generated (for both namespace and global typedef) so no issue but can't call alloc_type other than direct access in the input file.
The fix is to make castAt static so it's in translation unit scope and the linker never sees the duplicates.

### DIFF
--- a/include/trick/swig/cast_as.i
+++ b/include/trick/swig/cast_as.i
@@ -13,7 +13,7 @@
    conversion is used by the input processor to associate named allocations in the
    memory manager to names in pyton space.
  */
-TYPE * castAs##NAME( PyObject * obj1 ) {
+static TYPE * castAs##NAME( PyObject * obj1 ) {
     void *argp = NULL ;
     TYPE * ret = (TYPE *)NULL ;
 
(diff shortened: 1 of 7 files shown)


brendan-nasa:
Ah, looks like we just should just use the perltidy available on cpan (latest version) instead of whatever is in distro repos

hchen99:
> Ah, looks like we just should just use the perltidy available on cpan (latest version) instead of whatever is in distro repos

perltidy is anti-CI for sure ;-) or use a perltidy's configuration file .perltidyrc so the style preferences are controlled...but perltidy's internal formatting algorithms can still change between versions, producing different output even with the same config...or just give perl a break by either keep as is or only run perltidy locally ... we can discuss :-)

ddj116:
FYI from the ramtares perspective we think this MR is good to go. We have other issues we are working on but they are confirmed unrelated to this core issue. Thanks for y'alls help!

PR #2154, opened 2026-06-22
Add data recording support for STL (vector, deque, and array) of supported data types.

hchen99:
Add data recording support for STL (vector, deque, and array) of supported data types.
No .h5 file comparison for the test due to HDF5 files generated by DRHDF5.cpp embeds timestamps. md5sum comparison would fail on each run.


trick_source/sim_services/DataRecord/DRHDF5.cpp
brendan-nasa: this if check was and still isn't reachable?

hchen99: yea, should be outside of switch statement ;-)

PR #2164, opened 2026-07-21
clang setLangDefaults for clang >= 15

Willmac16:
I currently working on porting a project using trick from ubuntu 20 (clang 9) -> ubuntu 26 (clang 21) and was running into an issue with `//*` comments that worked on the old setup, but failed on the new one.

I found the commented out clang 15+ branch and set it up to preserve the prior clang behavior under the new namespace.


brendan-nasa:
What version of trick are you porting to?

And is your issue similar to https://github.com/nasa/trick/issues/2065 ?

Willmac16:
onto `25.1.0` so I believe I already have the PR #2077 that seemed to do the trick for that issue.

My interpretation is that we are not setting options that we otherwise would purely based on clang version.

sharmeye:
Thanks for bringing it up, we'll have to look into this.

sharmeye:
Setting those defaults for > 15 is fine and good practice.  Thanks for the contribution!

PR #2166, opened 2026-07-21
Not all init_attr routines are called at init.  #2165

### LINKED ISSUE #2165: Not all init_attr routines are called at init.
ICG builds class_map.cpp.  In the populate_class_map function I'd like to call the init_attr function.  In the past all of the init_attr functions that could be reached from the top level sim_objects were called.  This isn't a complete list.  This fix will address that.
 
c++
```
    // /users/alin/trick_sims/SIM_cfs_sbn/S_source.hh
    extern ATTRIBUTES  attrSieSimObject[] ;
    class_attribute_map->add_attr("SieSimObject" , attrSieSimObject) ;
    extern void init_attrSieSimObject() ;
    init_attrSieSimObject() ;
```

alexlin0:
If we are not generating attributes for sim_services, add a line to call each init_attr function as they are added to the attributes map.


sharmeye:
@alexlin0  Error: To fix, run: git clang-format master

your non-compliant code will no longer be tolerated!  put those braces on the next line!

sharmeye:
Looks great, where are the tests?

PR #2173, opened 2026-07-31
Fix GIL shutdown hang.

### LINKED ISSUE #2158: Trick 19.7.3 Simulation hangs at trick.exec_terminate()
A simulation based on Trick V19.7.3 running on RHEL 8.10 hangs at trick.exec_terminate(), and must be aborted terminated via Ctrl-c or kill command. The exec_terminate() (and similarly, exec_terminate_with_return()), seem to mostly just call ExecutiveException(). However, ExecutiveException() appears to be only setting the return code and outputting a string message. So, it is not clear why the routine should be hangin at this point?

What am I missing? Thanks for your help!

brendan-nasa:


include/trick/IPPython.hh
```diff
@@ -81,27 +71,14 @@ namespace Trick {
              @brief Restore variables with memory manager names to python space.
              @return always 0
             */
-            int restart() ;
-
-        protected :
-            /** TODO: remove units_conversion_msgs in 2021 */
-            /** false = see units conversion messages, true = head in sand */
-            bool units_conversion_msgs ;
```
brendan-nasa, line 89: >2021

trick_source/sim_services/InputProcessor/IPPython.cpp
```diff
@@ -66,9 +126,8 @@ void Trick::IPPython::get_TMM_named_variables() {
             ss << "trick.castAs" << user_type_name << "(int(" << alloc_info->start << "))" << std::endl ;
             ss << "except AttributeError:" << std::endl ;
             ss << "    pass" << std::endl ;
-            PyGILState_STATE gstate = PyGILState_Ensure();
-            PyRun_SimpleString(ss.str().c_str()) ;
-            PyGILState_Release(gstate);
+            GILGuard gil_guard;
+            PyRun_SimpleString(ss.str().c_str());
         }
```
sharmeye, line 131: maybe a comment saying that gil_guard needs to be destructed if you plan on using it again later in the same scope

PR #2174, opened 2026-08-01
Call deleteInstance before deleting oldType

stanbot8:
`AllocInfo::resize` destroys `oldType` before invoking `deleteInstance`.

Call `deleteInstance` before deleting `oldType`.

### DIFF
--- a/trick_source/sim_services/DataTypes/src/AllocInfo.cpp
+++ b/trick_source/sim_services/DataTypes/src/AllocInfo.cpp
@@ -215,8 +215,8 @@ void* AllocInfo::resize( size_t newElementCount ) {
         memcpy( newMemoryObject, oldMemoryObject, minByteCount);
     }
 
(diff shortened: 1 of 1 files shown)


brendan-nasa:
Thanks for the bug report (and fix)! In the interest of investigating further, did you encounter any crashes/run time errors that you traced back to these few lines?

stanbot8:
Hi Brendan, I was actually investigating a separate memory issue across my project's dependencies. I wrote a script that scans for suspicious patterns and this showed up alongside six other candidates. One of the other candidates turned out to be the culprit, but this looked worth submitting separately. I believe most Trick users won't encounter this issue, but the DataTypes test was updated last year so somebody might still be maintaining or using it.

sharmeye:
We've decided to shelve DataTypes for the time being until we circle back to include them in future work.  Thanks for the submission.

PR #2184, opened 2026-08-06
Support Swig 4.5.0

### LINKED ISSUE #2182: Trick will not compile for swig 4.5.0
While investigating another Trick bug, I had updated to the latest version of swig 4.5.0.

Trick is not compatible with the swig 4.5.0 API changes and I get Trick compile time errors.

The bigger issue is that there is no easy way to revert to an older version of swig on homebrew because "brew install swig@4.4" does not work. Homebrew does not provision for older versions of swig so you cannot revert. Anyone that updates homebrew swig will be broken until Trick fixes the compatibility issue.

brendan-nasa:
Swig 4.5.0 dropped support for Python 2. With that, they dropped compatibility `#define`s that allowed using the python 2 names for C API calls with Python 3.

This PR adds back the `#define`s that Trick uses.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,12 @@
 object_*
 io_src
 swig
+!include/trick/swig
+!share/trick/swig
+share/trick/swig/sim_services.py
+share/trick/swig/swig_double.py
+share/trick/swig/swig_int.py
+share/trick/swig/swig_ref.py
 lib
 lib64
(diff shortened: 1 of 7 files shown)


dandexter:
Thank you for working on a fix so quickly. I was able to successfully compile Trick and a simulation using swig 4.5.0.

PR #2190, opened 2026-08-19
refactor!: Remove dead and deprecated code.

brendan-nasa:
- **Remove _HAVE_TR1_RANDOM flag.** This flag was used to differentiate between draft C++11 (c++0x, [TR1](https://en.wikipedia.org/wiki/C%2B%2B_Technical_Report_1)) and standard C++11 compiler support. Neither of which is needed anymore because Trick's baseline C++ version is C++17.
- **Remove old compiler gate for unsupported c++ version (pre-C++11).**
- **`throw()` -> `noexcept`.** Replace empty `throw` expressions with `noexcept`. Destructors are implicitly `noexcept`
- **Remove old compiler gates. Bump minimum versions to gcc 8.5 and clang 14**. Removed gates that checked compiler versions older than our minimum (baseline minimum we require is gcc 8.5 and clang 14 from 2021 and 2022, respectively). `__GNUC__` is defined for both clang and gcc. Removed those checks.
- **Fix FindTrickICG**. Not really dead code but was incorrectly overriding the base class; particularly `Elif()` which we had defined as `ElIf()`. Wasn't caught because we used `virtual` instead of `override`.


share/trick/makefiles/Makefile.common
```diff
@@ -145,7 +145,6 @@ ifeq ($(IS_CC_CLANG), 1)
    TRICK_SYSTEM_CXXFLAGS += -fcolor-diagnostics
 else
    GCC_MAJOR := $(shell $(TRICK_CC) -dumpversion | cut -f1 -d.)
-   GCC_MINOR := $(shell $(TRICK_CC) -dumpversion | cut -f2 -d.)
```
sharmeye, line 148: are we sure we don't want to check the minor version, juuuuuust in case

PR #2195, opened 2026-08-29
fix(build): link ICG against clang component libs, not -lclang-cpp

brendan-nasa:
`-lclang-cpp` was originally added for Clang 16 because `clangAnalysis` began referencing `clangASTMatchers` through `UnsafeBufferUsage.cpp`. Rather than relying on `libclang-cpp`, detect `libclangASTMatchers.a` and `libclangAPINotes.a` by presence and add them in the required link order. This removes the Clang-version check and makes `-lclang-cpp` redundant across all supported Clang versions.

Verified by building `trick-ICG` with LLVM 14 through 23, along with a full build and `trick-CP` test on `SIM_cannon_analytic`.

`configure` regenerated with Autoconf 2.73.


sylvesterkaczmarek (commented):
The new extra-library detection only checks for `libclangASTMatchers.a` and `libclangAPINotes.a`, but the branch below explicitly supports `libclangFrontend.so` and therefore shared-component LLVM installs. In that configuration the required ASTMatchers/APINotes shared libraries are present as `.so` files, yet omitted from `OLD_CLANG_LIBS`, causing unresolved symbols. Please detect both static and shared forms (or probe linkability) and cover the shared-component path.

brendan-nasa:
This change only affects .a static archives

libClangAnalysis.so and libClangFrontend.so should carry ASTMatchers and APINotes.

Do you have a failing link line? or llvm install?

sylvesterkaczmarek:
I do not have a failing shared-library install or link line to point to. Re-reading the configure path, the extra ASTMatchers/APINotes entries are needed for the static component archives; when `libclangFrontend.so` / `libclangAnalysis.so` are selected, those internal dependencies are carried by the shared components. I was extrapolating from the static archive case rather than a reproduced shared-link failure, so I withdraw that concern.

PR #2213, opened 2026-09-17
fix: Interpret send_hs files using latin-1.

mbhatta:
fix: Interpret send_hs files using latin-1.

This can decode any byte value (0-255) into a character. Use the "backslashreplace" option which shows problematic bytes as \xbc, making them readable in stdout.


brendan-nasa:
try `encoding='utf-8', errors='surrogateescape'` instead

mbhatta:
> try `encoding='utf-8', errors='surrogateescape'` instead

@brendan-nasa Thanks for the suggestion, it works great.

brendan-nasa:
After looking at this closer, I think we can just parse the hs_file as raw bytes without worrying about decoding/encoding

share/trick/trickops/send_hs.py
```diff
@@ -102,30 +108,36 @@ def parse(self, hs_file=None):
...
-                      self._diagnostics[d] = self._attempt_hs_match(d, line)
-            # This text precedes the diagnostics output, use it as the trigger to start parsing
-            if 'REALTIME SHUTDOWN STATS:' in line:
-              start_reading_diagnostics = True
-              continue
+        # Specify encoding as utf-8. Use surrogateescape error handling to convert
+        # problematic bytes to surrogate characters that represent the original bytes.
+        # This will preserve all bytes from the file without data loss, even bytes
+        # that aren't valid utf-8.
+        with open(self.hs_file, encoding="utf-8", errors="surrogateescape") as fp:
```
brendan-nasa, line 126:
```suggestion
        with open(self.hs_file, "rb") as fp:
```

share/trick/trickops/send_hs.py
```diff
@@ -102,30 +108,36 @@ def parse(self, hs_file=None):
...
+            for line in fp:
+                self.num_lines += 1
+                #  NOTE this "attempt match" approach is less efficient but it should be
+                # robust to a future ordering change of the lines from Trick's output.
+                if start_reading_diagnostics:
+                    for d in self._diagnostics:
+                        if self._diagnostics[d] is None:
+                            self._diagnostics[d] = self._attempt_hs_match(d, line)
+                # This text precedes the diagnostics output, use it as the trigger to start parsing
+                if "REALTIME SHUTDOWN STATS:" in line:
```
brendan-nasa, line 138:
```suggestion
                if b"REALTIME SHUTDOWN STATS:" in line:
```

share/trick/trickops/send_hs.py
```diff
@@ -145,12 +157,12 @@ def _attempt_hs_match(self, name, text):
...
-        name = name.replace('(', '\(').replace(')', '\)')
-        m = re.match(name+': +([-]?[0-9]*\.?[0-9]+)', text.strip())
+        name = name.replace("(", "\(").replace(")", "\)")
+        m = re.match(name + ": +([-]?[0-9]*\.?[0-9]+)", text.strip())
         if m:
-            return(float(m.group(1)))
+            return float(m.group(1))
         else:
-            return(None)
+            return None
```
brendan-nasa, line 165:
```suggestion
        pattern = re.escape(name.encode("ascii")) + rb": +(-?[0-9]*\.?[0-9]+)"
        match = re.match(pattern, text.strip())
        return float(match.group(1)) if match else None
```

brendan-nasa (changes requested):

PR #2215, opened 2026-09-17
feat: Realtime Frame information

sharmeye:
Added variable server client and GUI to collect and display realtime frame performance information while a sim is running.


trick_sims/SIM_robot/S_define
brendan-nasa, line 1: was this supposed to be touched?

include/trick/JobData.hh
```diff
@@ -140,6 +140,9 @@ namespace Trick {
             /** time tic value from the executive */
             static long long time_tic_value ;      /**< trick_io(**) */
 
+            /** The total execution time of the job from the PREVIOUS frame (Stable for Variable Server) */
+            double prev_frame_time_seconds; /**< trick_units(s) */
```
brendan-nasa, line 144:
```suggestion
            double prev_frame_time_seconds {0}; /**< trick_units(s) */
```

trick_source/java/src/main/java/trick/RTJPerf/RealTimeJobPieChart.java
brendan-nasa, line 1: do you want `rtperf` or `rtjperf`

> `java/src/main/java/trick/RTJPerf` vs `trick.rtperf`

trick_sims/SIM_robot/S_define
sharmeye: it was my test sim, and i accidentally committed the changes then tried to un-change it but only kinda unchanged it.

trick_source/java/src/main/java/trick/RTJPerf/RealTimeJobPieChart.java
hchen99: package name to trick.rtjperf and the folder name to all lower case rtjperf too to be consistent with others?
also, TrickVariableServerClient maybe to RealTimeJobVariableServerClient as it's not a general purpose client, is rtjperf specific?

sylvesterkaczmarek (commented):
updateFrameData() divides every duration by totalDuration without handling zero. On an idle/first frame all job durations can be zero, so every percentage becomes NaN and then flows into the lists/chart. A zero-total frame needs an explicit path.

hchen99:
> updateFrameData() divides every duration by totalDuration without handling zero. On an idle/first frame all job durations can be zero, so every percentage becomes NaN and then flows into the lists/chart. A zero-total frame needs an explicit path.

A job is added to aggregatedDurations only when duration > 0.0 and duration < 100.0. frameData passed to updateFrameData() is populated from aggregatedDurations. If all jobs have 0 or invalid durations, frameData is empty. updateFrameData() returns immediately if newFrameData is empty. So, the totalDuration can't be 0 in updateFrameData(). The defensive guard is optional in this case.

sylvesterkaczmarek:
You're right. I missed the filtering boundary before updateFrameData: only durations > 0 and < 100 are inserted into aggregatedDurations, and an empty frame returns before the division. That makes totalDuration strictly positive whenever the percentage loop runs. My zero-total concern is withdrawn.

